PR #7, opened 2020-04-05
Add more fake functions for templates

chvck:
Added the ones that made sense to me to add. Notable exceptions are:

NumberWithFormat - I couldn't work out how to get from String to `&'static str`.
DateTimeBefore/After/Between - I couldn't figure out a day way to a) provide that in a nice way in a template and b)  use it in a template.
Paragraphs - Unsure what this is doing but it causes templates to freak out with:

```thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error("control character (\\u0000-\\u001F) found while parsing a string", line: 10, column: 0)', src/cli/fake_data/mod.rs:118:27```

### DIFF
--- a/src/cli/fake_data/mod.rs
+++ b/src/cli/fake_data/mod.rs
@@ -1,7 +1,16 @@
 use super::util::convert_json_value_to_nu_value;
 use crate::state::State;
+use fake::faker::address::raw::*;
+use fake::faker::boolean::raw::*;
+use fake::faker::chrono::raw::*;
+use fake::faker::company::raw::*;
+use fake::faker::currency::raw::*;
+use fake::faker::filesystem::raw::*;
 use fake::faker::internet::raw::*;
+use fake::faker::lorem::raw::*;
 use fake::faker::name::raw::*;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/cli/fake_data/mod.rs [hunk 1]
```diff
@@ -186,6 +195,197 @@ fn register_functions(tera: &mut Tera) {
...
+                Err(_) => {
+                    return Err(tera::Error::msg(format!(
+                        "Function `numberWithFormat` received format={} but `format` can only be a string",
+                        val
+                    )));
+                }
+            },
+            None => String::from("#"),
+        };
+        let format_str = Box::leak(format.into_boxed_str());
```
daschl, line 216: I think we should add a comment saying that we know this "leaks" memory into the running program, but it should hopefully be called once and it is tiny. But since it's technically still a leak...

daschl:
@chvck looks great! Just add the comment and then feel free to merge

daschl:
@chvck oh and maybe update the commit msg to not include the failures - we can file tickets for them if we want to and investigate separately

PR #8, opened 2020-04-14
Wire up and update support for kv-upsert

chvck:


src/cli/upsert.rs
```diff
@@ -78,26 +111,36 @@ async fn run_upsert(
...
+
+        // An empty value is a legitimate document
+        rows.insert(arg_id, arg_content);
     }
 
     let bucket = state.active_cluster().cluster().bucket("travel-sample");
     let collection = bucket.default_collection();
 
-    debug!("Running kv get for docs {:?}", &ids);
+    debug!("Running kv get for docs {:?}", &rows);
```
daschl, line 132: Running kv upsert ..

src/cli/upsert.rs
```diff
@@ -78,26 +111,36 @@ async fn run_upsert(
...
 
     let mut results = vec![];
-    for id in ids {
-        match collection.get(&id, GetOptions::default()).await {
-            Ok(res) => {
+    for (id, content) in rows.iter() {
+        match collection
+            .upsert(id, content, UpsertOptions::default())
```
daschl, line 137: I see you are storing the content as a string, but right now I'm not sure if the upsert thinks (simple code in couchbase-rs) that is a string type and will double encode it? Maybe we need to (even if wasteful) for now turn it into a JsonValue with serde again before inserting.. if it looks fine on the server then all good

PR #11, opened 2020-04-18
Add support for KV insert

chvck:


src/cli/insert.rs
```diff
@@ -0,0 +1,105 @@
...
+        .get("id-column")
+        .map(|id| id.as_string().unwrap())
+        .unwrap_or_else(|| String::from("id"));
+
+    let content_column = args
+        .get("content-column")
+        .map(|content| content.as_string().unwrap())
+        .unwrap_or_else(|| String::from("content"));
+
+    let mut rows = json_rows_from_input_columns(&mut args, &id_column, &content_column).await?;
```
chvck, line 79: I'm not totally confident that id_column and content_column need to be borrows but it might make sense in other call cases?

daschl: if it compiles it should be fine, good rule of thumb with rust ;) ..

PR #41, opened 2020-07-22
Continue hardening commands

chvck:


src/cli/whoami.rs
```diff
@@ -62,7 +62,10 @@ async fn whoami(
         .map(|id| id.as_string().unwrap())
         .unwrap_or_else(|| state.active());
 
-    let cluster_identifiers = cluster_identifiers_from(&state, identifier_arg.as_str());
+    let cluster_identifiers = match cluster_identifiers_from(&state, identifier_arg.as_str()) {
```
daschl, line 65: can't use just ? here like in the buckets command?

PR #60, opened 2020-10-19
If cli options are set then override "default" config cluster settings

chvck:


src/main.rs
```diff
@@ -34,55 +34,100 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+        true => Some(rpassword::read_password_from_tty(Some("Password: ")).unwrap()),
+        false => None,
     };
 
     let active = if config.clusters().is_empty() {
         let timeouts = ClusterTimeouts::default().export_lcb_args();
+        let hostnames = if let Some(hosts) = opt.hostnames {
+            hosts
+        } else {
+            "localhost".into()
```
daschl, line 46: nit: maybe make localhost, administrator etc const values at the top of the file? makes it easier to change/reuse later

daschl (approved):
just a minor nit, can even be done later. feel free to merge at any point

daschl (approved):

PR #64, opened 2020-10-22
Add support for collections to doc commands

chvck:


src/cli/use_cmd.rs
```diff
@@ -56,6 +56,20 @@ async fn use_cmd(
             .map(|s| s.clone())
             .unwrap_or(String::from("<not set>")),
     );
+    using_now.insert_value(
+        "scope",
+        active
+            .active_scope()
+            .map(|s| s.clone())
+            .unwrap_or(String::from("<not set>")),
```
daschl, line 64: since we kinda can infer the default collection and scope to _default, should we assume that instead of not set?

daschl (approved):

PR #77, opened 2020-12-01
Initial stab at a testing framework

chvck:


tests/common/playground.rs
```diff
@@ -0,0 +1,123 @@
...
+struct State {
+    connstr: String,
+    bucket: String,
+    scope: String,
+    collection: String,
+    username: String,
+    password: String,
+}
+
+pub struct CBPlayground {}
```
chvck, line 46: Once we need it we can add nu playground as a property of this. I did have it that way but dead_code complained at me.

daschl (approved):

PR #81, opened 2020-12-18
[breaking] change cluster config to list from map

daschl:
This change simplifies the format and makes it easier
to provide dot-separated identifiers.


chvck (commented):
I think that the README will need updating too.

chvck (approved):

PR #82, opened 2021-01-13
Update nushell version so we can compile with latest rust release

chvck:


src/main.rs
```diff
@@ -245,7 +245,7 @@ async fn main() -> Result<(), Box<dyn Error>> {
     ]);
 
     if let Some(c) = opt.command {
-        nu_cli::run_pipeline_standalone(c, opt.stdin, &mut context, true).await?;
+        nu_cli::run_script_file(c, true).await?;
```
chvck, line 248: Less than confident these are the correct commands.

daschl: from the naming it makes sense

daschl (approved):

PR #90, opened 2021-01-30
avoid cookie JS popup when in an iframe

steveyen:
As part of integrating a couchbase-shell session into an web page in the Couchbase Playground, I figured having its docs conveniently right on the same page would be nice, which I attempted via an old-school iframe.

It works.

Except one annoyance... the popup to accept cookies appears every single time in the iframe.  See screenshot.  
Screen Shot 2021-01-30 at 9 06 36 AM

So, this fix is to avoid the "accept cookies" popup when the couchbase-shell docs is in an iframe.  The outermost web pages already should have made the user go through the Accept Cookies dance steps.

NOTE - I couldn't figure out how to locally test this.

But, I'm thinking the downside risk of this change is that it will disable the cookie popup (perhaps with rejoicing?) ...and in which case this change ought to be easy to revert.


daschl (approved):

daschl:
pushed a new build to the gh-pages branch - I hope it works now?

PR #92, opened 2021-02-03
Log sdk output to file and add an sdklog command

chvck:
Kind of POC, it works pretty well. Redirects all SDK output to a file so that it never floods the terminal. Also added a command to get the last x log lines although I'd expect most users would just use tail or the like.


src/cli/sdk_log.rs
```diff
@@ -0,0 +1,59 @@
...
+    let last = match args.get("last") {
+        Some(v) => match v.as_u64() {
+            Ok(l) => l,
+            Err(e) => return Err(e),
+        },
+        None => 10,
+    };
+
+    debug!("Fetching last {} lines from sdk log", &last);
+    let mut file =
```
daschl, line 43: should this handle a non-existent file gracefully instead of panicing?

src/main.rs
```diff
@@ -25,7 +25,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
     const DEFAULT_HOSTNAME: &str = "localhost";
     const DEFAULT_USERNAME: &str = "Administrator";
 
-    pretty_env_logger::init();
+    // pretty_env_logger::init();
```
daschl, line 28: leftover?

src/main.rs
```diff
@@ -25,7 +25,8 @@ async fn main() -> Result<(), Box<dyn Error>> {
...
+    log4rs::init_file("log4rs.yml", Default::default()).unwrap();
```
daschl, line 29: what happens if that file is not present? if you cargo run from the dir  it is there, but once we ship the distribution it might be not -- should we include a default config as part of the binary and optionally check for that logfile somewhere?

daschl (approved):
that's a great improvement

PR #104, opened 2021-03-09
Logdir

chvck:


src/cli/tutorial.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn signature(&self) -> Signature {
+        Signature::build("tutorial").optional(
+            "action",
+            SyntaxShape::String,
+            "the action to perform - one of next or prev",
+        )
+    }
+
+    fn usage(&self) -> &str {
+        "Perform document operations against a bucket or collection"
```
daschl, line 34: is that usage copied from somwhere? Maybe that it is a tutorial might be a better description? (since it might also introduce more concepts around nushell etc)

src/tutorial.rs
```diff
@@ -0,0 +1,143 @@
...
+        let mut current_step = self.current_step.lock().unwrap();
+        *current_step -= 1;
+        if *current_step < 0 {
+            *current_step = self.num_steps - 1;
+        }
+        STEPS[*current_step as usize].clone()
+    }
+}
+
+static STEPS: &'static [&'static str] = &[
```
daschl, line 42: what if we make this a Map of str str ? The name could then be used to list all the steps in a table when doing something like "tutorial steps" ? .. I could imagine someone want to jump to a specific tutorial step like i.e. query or something

daschl (approved):

src/tutorial.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    cat output.csv
+
+You can try the same for other formats too, such as JSON.
+Try 'to --help' to see the available formats.
+
+Use 'tutorial next' to move to the next step in the tutorial.
+    ");
+        m.insert(
+            "Page conclusion",
```
chvck, line 168: Page ... doesn't work

PR #132, opened 2021-11-17
Perform kv operations in batches of up to 500

chvck:


src/client/kv_client.rs
```diff
@@ -27,49 +31,94 @@ impl KvResponse {
     pub fn cas(&self) -> u64 {
         self.cas
     }
+
+    pub fn id(&self) -> String {
```
daschl, line 35: since this is all internal, why use both id and key? (maybe just settle on one?)

src/client/kv_client.rs
```diff
@@ -27,49 +31,94 @@ impl KvResponse {
...
+                password.clone(),
+                bucket.clone(),
+                tls_config.clone(),
+            );
+
+            let endpoint = select! {
+                res = connect => res,
+                () = &mut deadline_sleep => Err(ClientError::Timeout{key: None}),
+                () = &mut ctrl_c_fut => Err(ClientError::Cancelled{key: None}),
+            }?;
```
daschl, line 96: since this will back out if one endpoint times out - what happens to the ones already connected? does the Drop handle the disconnect for them automatically?

src/client/kv_client.rs
```diff
@@ -88,50 +142,40 @@ impl KvClient {
...
             let host: String;
             let port: i32;
             if host_split.len() == 1 {
                 host = seed.clone();
-                port = if self.tls_config.enabled() {
-                    18091
-                } else {
-                    8091
-                };
+                port = if tls_config.enabled() { 18091 } else { 8091 };
```
daschl, line 166: so we have custom port support yet?

src/cli/doc_get.rs
```diff
@@ -112,6 +114,25 @@ fn run_get(state: Arc<Mutex<State>>, mut args: CommandArgs) -> Result<OutputStre
         ids.push(id);
     }
 
+    let batch_size = 500;
```
daschl, line 117: can we make that a configurable / static somewhere? so it's not duplicated for each command (i.e. in the user config somewhere)

src/cli/doc_insert.rs
```diff
@@ -85,136 +78,12 @@ impl nu_engine::WholeStreamCommand for DocInsert {
     }
 }
 
-fn run_insert(state: Arc<Mutex<State>>, args: CommandArgs) -> Result<OutputStream, ShellError> {
```
daschl, line 88: nice cleanup

src/cli/doc_get.rs
chvck: I was going to do that as a later improvement, it should be a user configurable setting - along with max number kv connections when/if we add that.

src/client/kv_client.rs
chvck: I've not actually looked into if this will terminate the connection but I would assume that the connection future will stop being polled and it'll drop the KvEndpoint triggering disconnect.

src/client/kv_client.rs
chvck: I *think* that this if is the custom port support. Check out the else below. This block covers the case where the user didn't define a port so we use the defaults here.

PR #163, opened 2022-03-16
Update to nushell 0.59.1

chvck:


Cargo.toml
```diff
@@ -21,21 +22,26 @@ hmac = "0.11.0"
...
-nu-data = { version = "0.42.0" }
-nu-errors = { version = "0.42.0" }
-nu-stream = { version = "0.42.0" }
-nu-source = { version = "0.42.0" }
-nu-cli = { version = "0.42.0" }
-nu-engine = { version = "0.42.0" }
-nu-plugin = { version = "0.42.0" }
-nu_plugin_from_bson = { version = "0.42.0" }
-nu-value-ext = { version = "0.42.0" }
+nu = { path = "../nushell", features = ["extra"] }
```
chvck, line 29: These will obviously need to change, once my changes are merged upstream.

src/cli/cbenv_bucket.rs
```diff
@@ -0,0 +1,64 @@
...
+
+impl UseBucket {
+    pub fn new(state: Arc<Mutex<State>>) -> Self {
+        Self { state }
+    }
+}
+
+impl Command for UseBucket {
+    fn name(&self) -> &str {
+        "cb-env bucket"
```
chvck, line 23: Open to better names than cb-env but given other nushell commands I think it makes sense.

Cargo.toml
ingenthr: I think that's done now, right?  Need to update this PR?

chvck: The PR has been updated already, it was using local paths but is now using a git SHA. Nushell 0.60 is out now so once this PR is in we can make a new patchset to bump to 0.60 and use a proper tagged version again.

src/cli/cbenv_cmd.rs
```diff
@@ -0,0 +1,163 @@
...
+                    );
+                    results.add_i64(
+                        "search-timeout (ms)",
+                        timeouts.search_timeout().as_millis() as i64,
+                        span,
+                    );
+                }
+            }
+            None => {
+                results.add_string("username", String::from("<not set>"), span);
```
ingenthr, line 153: similar comment here.  seems like the value should be some kind of optional and what you have here is really formatting.

src/cli/cbenv_project.rs
```diff
@@ -0,0 +1,57 @@
...
+    ) -> Result<PipelineData, ShellError> {
+        let guard = self.state.lock().unwrap();
+        let active = guard.active_capella_org()?;
+
+        active.set_active_project(call.req(engine_state, stack, 0)?);
+
+        let mut result = NuValueMap::default();
+        result.add_string(
+            "project",
+            active.active_project().unwrap_or_else(|| String::from("")),
```
ingenthr, line 51: was this intentionally different?

ingenthr (approved):
I'm going to do some additional looking around here and possibly change my review.

I have some comments that might be more improvements.  I don't think they should block merge, but they may just be things we can improve in the configuration type system so to speak.

src/cli/cbenv_bucket.rs
chvck: >  Are there popular extensions

No :)

docs/cloud.adoc
```diff
@@ -9,7 +9,7 @@ When using <<_management_commands>> with Couchbase Capella and Couchbase Shell w
 You can use the `use --capella` command to see which cloud resources are current active:
 
 ```
-> use --capella
```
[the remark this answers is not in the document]
chvck: `use` is now a nushell language keyword. It's not even a command in the way that something like `ls` is or we could make a workaround for it.

src/cli/buckets_builder.rs
```diff
@@ -41,7 +42,10 @@ impl TryFrom<&str> for DurabilityLevel {
             "majority" => Ok(DurabilityLevel::Majority),
             "majorityAndPersistActive" => Ok(DurabilityLevel::MajorityAndPersistOnMaster),
             "persistToMajority" => Ok(DurabilityLevel::PersistToMajority),
-            _ => Err(ShellError::unexpected("invalid durability mode")),
+            _ => Err(generic_labeled_error(
+                "invalid durability mode",
```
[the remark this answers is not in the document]
chvck: We definitely need to go through and improve errors in general to improve UX, there is an issue open for that.

src/cli/cbenv_bucket.rs
```diff
@@ -0,0 +1,64 @@
...
+        };
+
+        active.set_active_bucket(call.req(engine_state, stack, 0)?);
+
+        let mut result = NuValueMap::default();
+        result.add_string(
+            "bucket",
+            active
+                .active_bucket()
+                .unwrap_or_else(|| String::from("<not set>")),
```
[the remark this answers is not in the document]
chvck: This is just the name to display within the table, it's not actually ever used in code as such.

src/cli/cbenv_project.rs
chvck: An oversight I think.

PR #181, opened 2022-06-21
Automatically detect query context

chvck:


chvck:
I've made this so that it will always run with a query context if an active scope is set (or the scope flag used) and `disable-context` flag not used.  Sounds about right @ebenhaber @ingenthr ?

ingenthr:
Yes, that sounds correct to me.  Thanks!

PR #194, opened 2022-08-23
Removevpc

chvck:


chvck:
I've been pretty militant in removing dead code here. Seems like a better idea to add things back as required.

PR #201, opened 2022-11-14
Update to nu 0.71.0

chvck:


brett19 (changes requested):
Generally speaking the commit looks good, but looks like lots of build errors.

brett19:
Side-note: Why do the Checks show as passing, but there are a ton of errors reported?

chvck:
> Side-note: Why do the Checks show as passing, but there are a ton of errors reported?

The build is passing and so is cargo fmt. Cargo clippy is way a more picky linter and right now I'm using it for information rather than a build failure.

brett19 (approved):

PR #202, opened 2022-11-17
Updates to make clippy happy

chvck:


brett19 (approved):
Looks good, although windows is complaining about an unused std::fs import.

PR #207, opened 2022-11-30
Improve doc command error messages

chvck:
Rework client package to not have references to nushell, making wrapping any ClientErrors the responsibility of the cli package. Add message() and expanded_message() to ClientError to facilitate conversion from ClientError to ShellError.
Improve the messages and error details for ClientError. Add a halt-on-error flag to doc commands to facilitate scripting where error codes might be wanted.

### DIFF
--- a/src/cli/analytics.rs
+++ b/src/cli/analytics.rs
@@ -1,5 +1,6 @@
 use crate::cli::error::{
-    deserialize_error, malformed_response_error, unexpected_status_code_error,
+    client_error_to_shell_error, deserialize_error, malformed_response_error,
+    unexpected_status_code_error,
 };
 use crate::cli::util::{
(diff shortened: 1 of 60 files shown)


chvck:
This... got out of hand. What I'm primarily interested in here is feedback on the idea that commands in cbshell will always return an "error" column in results (same as there's a "cluster" column added), which will contain the abridged error(s) for the given key/cluster as applicable. 

In the doc commands I've also added a flag override --halt-on-err (-e) which means that on first error encountered the command stops and will return the full error (including help) on stderr with an appropriate error code (that bit is handled by nushell) .

chvck:
Having discussed with Brett and thought about this a bit more I'm not going to extend halt-on-error to other commands. I think that it's still important for doc commands because of "doc not found"/"doc exists" and errors of that nature. I will improve the messages for other commands but they will always error on first error.

PR #208, opened 2022-12-21
Update to nu v0.73.0

chvck:


chvck:
This change is actually quite small really. The changes in help.rs are because I've updated our help with a direct c&p from nushell where we just update the base help message. I need to make a change in nushell and upstream it at some point so that we don't have to do this.

The other primary change is to the prompt creation - changed to use "and" instead of && and to use + to concatenate strings rather than using build-string. Both && and build-string no longer exist in nu.

brett19 (approved):

PR #214, opened 2023-02-09
Cargo update all dependencies

chvck:


ingenthr (approved):
Hard to review this, but LGTM.  Newer is better, right?

PR #219, opened 2023-03-07
Update version to 0.75.0

chvck:


Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "couchbase-shell"
-version = "1.0.0-beta.5"
+version = "0.75.0"
 authors = ["Michael Nitschinger <michael@nitschinger.at>", "Charles Dixon <chvckd@gmail.com>"]
```
ingenthr, line 4: Should we update this and the next line?

ingenthr (approved):
Might want to update some of the other package metadata.

Cargo.toml
chvck: Updating the edition should be its own commit, it'll require a bunch of testing etc... as it's referring to the underlying rust language compatibility and could break some things that will need fixing.

ingenthr: ah, ok.  that's fine then.

PR #222, opened 2023-03-23
Cluster type detection

chvck:


src/remote_cluster.rs
```diff
@@ -0,0 +1,237 @@
...
+pub struct ClusterTimeouts {
+    data_timeout: Duration,
+    query_timeout: Duration,
+    analytics_timeout: Duration,
+    search_timeout: Duration,
+    management_timeout: Duration,
+}
+
+impl Default for ClusterTimeouts {
+    fn default() -> Self {
```
ingenthr, line 170: This makes me wonder, do we have configuration profiles here?

ingenthr (approved):

src/remote_cluster.rs
chvck: No we don't. I think we easily could do in the future though. We could even automatically switch the default profile depending on cluster type. https://github.com/couchbaselabs/couchbase-shell/issues/228 filed.

PR #223, opened 2023-03-23
Rename commands and update references

chvck:
This is dependent on the config_rename PR going in first.


src/cli_options.rs
```diff
@@ -0,0 +1,344 @@
...
+    Value,
+};
+use std::io::Write;
+
+#[derive(Clone, Debug)]
+pub struct CliOptions {
+    pub conn_string: Option<String>,
+    pub username: Option<String>,
+    pub password: bool,
+    pub cluster: Option<String>,
```
ingenthr, line 17: so, I guess these stay the same in the move to database from cluster?

ingenthr (approved):
LGTM.  Made a note on something to look at, but it could be handled in the future if required.

src/cli_options.rs
chvck: These are just struct members used internally. Parsing into them happens below. You're right that cluster need renaming to database in the parsing too though.

chvck: Filed #229

PR #225, opened 2023-03-23
Update tutorial

chvck:


docs/reference.adoc
```diff
@@ -9,9 +9,9 @@ The `~/.cbsh/config` file with examples:
 # Allows us to evolve in the future without breaking old config files
 version = 1
 
-[[cluster]]
+[[database]]
 identifier = "default"
-hostnames = ["127.0.0.1"]
+conn-string = "127.0.0.1"
```
ingenthr, line 14: probably a bad time to bring this up, but "conn-string" isn't my preference.  why a dash here?  I tend to use "connstr" in most (or all) places.

ingenthr (approved):
LGTM.  Made a note for you to consider, but we can do that after all of these go in.

docs/reference.adoc
chvck: Filed #230

PR #226, opened 2023-03-23
Capella certs

chvck:


src/client/kv.rs
```diff
@@ -62,7 +62,13 @@ impl KvTlsConfig {
                     key: None,
                 })?
             } else {
-                debug!("Adding Capella root CA to trust store");
+                for cert in rustls_native_certs::load_native_certs()
+                    .expect("Could not load native platform certs")
```
ingenthr, line 66: Do we also have the self-signed for Capella Provisioned?  Both the prod and nonprod environments?

ingenthr (approved):

src/client/kv.rs
chvck: Just below is `let mut reader = BufReader::new(CAPELLA_CERT.as_bytes());` which is for the prod env. We do not have the self-signed for non-prod as far as I'm aware.

PR #237, opened 2023-03-29
Rename conn-string to connstr

chvck:


ingenthr (approved):
I like this change.  That said, if you feel conn-string is more clear or we take a poll and others do, we can keep it.  

image

image

chvck:
> I like this change. That said, if you feel conn-string is more clear or we take a poll and others do, we can keep it.
> 
> There's some evidence though that we do it the other way:
> 
> image
> 
> image

connstr works for me, I think that it's pretty clear.

PR #242, opened 2023-04-13
Shorten the windows prompt

chvck:
For some reason it being longer is causing a stack overflow.


ingenthr (approved):
Weird issue.  Did we file upstream?  

This change looks good to me.

chvck:
> Weird issue. Did we file upstream?
> 
> This change looks good to me.

It doesn't seem to be an upstream issue. I can repro this on cbshell with non-cbsh commands but can't on Nushell. Will investigate more.

PR #244, opened 2023-04-14
Update docs

chvck:


docs/quickstart.adoc
```diff
@@ -27,35 +28,42 @@ You can now run the `cbsh` binary:
 The Couchbase Shell 0.75.1
 ```
 
-TIP: If you are running a recent macOS release (i.e. 10.15.x), you'll likely see an error similar to *"cbsh" was blocked from use because it is not from an identified developer*. This is because our binaries are not yet signed. To run it nonetheless you need to navigate to `System Preferences -> Security & Privacy` and click `Allow Anyway`. Next time you run the binary you'll get another prompt but then it should run fine. 
+TIP: If you are running a recent macOS release (i.e. 10.15.x), you'll likely see an error similar to *"cbsh" was blocked from use because it is not from an identified developer*.
+This is because our binaries are not yet signed.
+To run it nonetheless you need to navigate to `System Preferences -> Security & Privacy` and click `Allow Anyway`.
+Next time you run the binary you'll get another prompt but then it should run fine.
```
ingenthr, line 34: This is also possible at the CLI with
```
sudo xattr -r -d com.apple.quarantine $PWD/cbsh                          
```

ingenthr (approved):
I have two things that could be improvements, but fine for this to go forward now.  Looks like you did a lot of work on this!  Thanks for all the help.

PR #249, opened 2023-04-19
Prepare v0.75.2 release

chvck:


chvck:
Purposefully not updating docs or README as this release doesn't actually add much value from a user POV. It's primarily to try unblock homebrew for us.

brett19 (approved):

PR #252, opened 2023-04-30
Add a Dockerfile

brantburnett:
Motivation
----------
Allow generation of Docker image which may be used within Kubernetes to manage Couchbase clusters where there is no network access from the developer's PC.

Modifications
-------------
Add a multi-stage Dockerfile.

Results
-------
A Docker image may be easily generated. This doesn't implement any CI/CD or publishing of the Docker image to Docker Hub.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,10 @@
+.git/
+target/
+log/
+.vscode/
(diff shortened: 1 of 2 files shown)


chvck:
Awesome, thanks @brantburnett

PR #257, opened 2023-07-31
Update password flag to be more idiomatic

chvck:


chvck:
I don't love this but also cannot find another approach.

ingenthr (commented):
So, the canonical way to do this in a CLI tool is to have two options:

Ask me for the password at stdin:
```
-p -
```

Use the string "foobar" as a password:
```
-p foobar
```

Then you can require -p as an argument and if it's elided, error out.  

Seems like it'd be a better approach.  Also, it's sometimes nice to be able to specify password is in a file.  This is a security consideration, since on UNIX users can see the command line flags from other users.

ingenthr:
Related: https://www.netmeister.org/blog/passing-passwords.html

src/cli_options.rs
```diff
@@ -58,9 +58,10 @@ impl Command for Cbsh {
                 "name to show in the shell",
                 None,
             )
-            .switch(
+            .named(
                 "password",
-                "use to specify a password to use for authentication",
+                SyntaxShape::String,
+                "use to trigger a prompt to specify a password to use for authentication",
```
ingenthr, line 64: I think with the change, this line isn't quite right, is it?  It's provide the password or prompt it `-` is specified.

ingenthr (changes requested):
Note on the help there…

ingenthr (approved):

PR #261, opened 2023-08-31
Change TLS trust store

chvck:
Motivation
----------
We are currently using the native store to loads certs for use with kv. The native package has a dependency on OpenSSL which we don't want. We are allowing reqwest to load its default trust store that it uses for rusttls - webpki-roots. We should change both of these so that we load webpki-roots ourselves and set the config to use on reqwest - so that we control it and have better tracibility into versions being used as well as consistency across both HTTP and KV.

Changes
-------
Refactor TLS throughout so that TLS configs are only setup once per cluster entry in the config.
Refactor TLS so that HTTP and KV use the same TLS object type. Always uses webspki-roots as our trust store.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -48,7 +48,6 @@ reqwest = { version = "0.11", features = ["json", "rustls-tls", "blocking"] }
 rpassword = "7.0.0"
 rust-embed = "6.3"
 rustls-pemfile = "1.0.0"
-rustls-native-certs = "0.6.2"
 serde = "1.0"
 serde_json = "1.0"
 serde_derive = "1.0"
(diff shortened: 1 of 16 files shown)


chvck:
I'm open to changing this to use the native TLS store instead of the Mozilla store. Consistency across the entire shell and us having control is what matters more IMO.

Cargo.toml
```diff
@@ -79,8 +79,8 @@ nu-test-support = { version = "0.83.1"}
 strum = "0.24.1"
 strum_macros = "0.24.3"
 
-[features]
 
+[features]
 # Enable to statically link OpenSSL; otherwise the system version will be used. Not enabled by default because it takes a while to build
 static-link-openssl = ["dep:openssl"]
```
ingenthr, line 85: If we don't need it any longer, why do we have it here?

ingenthr (approved):
One question in there, but approving assuming that it's just my misunderstanding.  Let me know if you need an updated review.

Cargo.toml
chvck: We need it on Unix systems (not macos though). Nushell itself uses a crate which builds openssl, so on Linux we statically link at CI time.

PR #265, opened 2024-01-29
Add command to get organizations using v4 endpoint

Westwooo:
WIP


src/cli/organizations.rs
```diff
@@ -0,0 +1,100 @@
...
+    _input: PipelineData,
+) -> Result<PipelineData, ShellError> {
+    let span = call.head;
+    let ctrl_c = engine_state.ctrlc.as_ref().unwrap().clone();
+
+    debug!("Running organizations");
+
+    let guard = state.lock().unwrap();
+    let control = guard.active_capella_org()?;
+    let client: Arc<crate::client::CapellaClient> = control.client();
```
brett19, line 67: I think this might be unnecessarily specifying the type?

src/cli/organizations.rs
```diff
@@ -0,0 +1,100 @@
...
+    let response = client
+        .capella_request(
+            CapellaRequest::GetOrganizations {},
+            Instant::now().add(control.timeout()),
+            ctrl_c,
+        )
+        .map_err(|e| client_error_to_shell_error(e, span))?;
+
```
brett19, line 75: nit: This newline doesn't match the projects.rs formatting, I'm not precisely sure about the code-style here, but comparing the two files, this looked different.

src/client/cloud.rs
```diff
@@ -84,12 +84,16 @@ impl CapellaClient {
         mac.update(bearer_payload.as_bytes());
         let mac_result = mac.finalize();
 
-        let bearer = format!(
+        let mut bearer = format!(
```
brett19, line 87: Rather than changing this to mut, and then generating a token, just to overwrite it, why not just generate the appropriate kind of bearer token for the API that's in use directly?

brett19 (changes requested):

src/cli/organizations.rs
Westwooo: You're right, done

brett19 (approved):

PR #266, opened 2024-01-30
use v4 API for projects commands

Westwooo:


src/cli/databases_create.rs
```diff
@@ -105,7 +124,7 @@ fn clusters_create(
                 payload: serde_json::to_string(&json)
                     .map_err(|e| serialize_error(e.to_string(), span))?,
             },
-            Instant::now().add(control.timeout()),
+            Instant::now().add(timeout),
```
brett19, line 127: Should this be using deadline rather than timeout?

src/cli/databases_create.rs
```diff
@@ -78,22 +79,40 @@ fn clusters_create(
...
             return Err(no_active_project_error(span));
         }
     };
-    let project_id = find_project_id(ctrl_c.clone(), project_name, &client, deadline, span)?;
+
+    let org_id = match control.id() {
+        Some(id) => id,
+        None => {
+            let id = find_org_id(ctrl_c.clone(), &client, deadline, span)?;
+            guard.set_active_capella_org_id(id.clone())?;
```
brett19, line 103: Why do we set global state here for a local command specifying the id?

brett19 (changes requested):

src/cli/databases_create.rs
Westwooo: So to do anything with projects we need the org_id. Instead of having to fetch it every time we do a command I wanted to save it for future use, so if the org_id is blank then we fetch it and add it to the struct representing the org.

src/cli/databases_create.rs
Westwooo: Yes it should, done

src/state.rs
```diff
@@ -191,6 +188,25 @@ impl State {
         Ok(())
     }
 
+    pub fn set_active_capella_org_id(&mut self, id: String) -> Result<(), ShellError> {
+        let active = self.active_capella_org_name().unwrap();
```
chvck, line 192: Can this not panic? If it can't be empty then why does the func return an option?

Westwooo: You're right, done now

PR #267, opened 2024-02-05
Add support for testing against a Capella cluster

Westwooo:


tests/integration/util/standalone.rs
```diff
@@ -33,13 +33,18 @@ impl StandaloneCluster {
...
         let username = c.username().unwrap();
         let password = c.password().unwrap();
-        let features = StandaloneCluster::get_supported_features(
-            conn_str.clone(),
-            bucket.clone(),
-            username.clone(),
-            password.clone(),
-        )
-        .await;
+        let features = if conn_str.contains("couchbases") {
```
chvck, line 36: This seems weird to me, doesn't this mean that we can't test with SSL against a non-capella cluster?

tests/integration/util/features.rs
```diff
@@ -19,6 +20,7 @@ impl From<&str> for TestFeature {
             "queryindex" => TestFeature::QueryIndex,
             "queryindexdefs" => TestFeature::QueryIndexDefinitions,
             "queryindexadvise" => TestFeature::QueryIndexAdvise,
+            "capella" => TestFeature::Capella,
```
chvck, line 23: I'm unsure that this is the correct way to approach this. Couldn't we use the presence of the access/secret keys to indicate capella support?

tests/integration/util/config.rs
```diff
@@ -17,6 +17,11 @@ pub struct Config {
     bucket: Option<String>,
     enabled_features: Vec<TestFeature>,
     data_timeout: String,
+    capella_conn_string: Option<String>,
+    capella_username: Option<String>,
```
chvck, line 21: Do we need to duplicate username and password, can we not just use the existing config properties for that? Could we maybe add a new cluster_type for capella or something?

PR #269, opened 2024-02-16
Vector generate command

Westwooo:
The vector enrich-text command is used to produce vector indexable json docs form local data. The JSON will be of the form: 
```
{
     "id": "vectorx",
     "text":  "some chunk of text data",
     "vector": [0.187987123, ...., -0.0128379],
} 
```
It takes either a list of files output by ls:
```
ls | vector enrich-text
```

Or a string, this can be the contents of a single file, or user defined:
```
"embed this for me please" | vector enrich-text
open some-file.txt | vector enrich-text
```

The JSON output can be written directly to Couchbase: 
```
open ../some-file.txt | vector enrich-text | doc upsert
```

### DIFF
--- a/src/cli/util.rs
+++ b/src/cli/util.rs
@@ -354,6 +354,34 @@ pub fn duration_to_golang_string(duration: Duration) -> String {
     golang_string
 }
 
+pub fn read_openai_api_key(engine_state: &EngineState) -> Result<String, ShellError> {
+    let key = match engine_state.get_env_var("OPENAI_API_KEY") {
+        Some(k) => match k.as_string() {
+            Ok(k) => k,
+            Err(e) => {
+                return Err(ShellError::GenericError(
+                    format!("could not read OPENAI_API_KEY env var as a string: {}", e),
+                    "".to_string(),
+                    None,
+                    None,
+                    Vec::new(),
+                ));
+            }
+        },
+        None => {
+            return Err(ShellError::GenericError(
+                "Please specify API key using: \"$env.OPENAI_API_KEY = <YOUR API KEY>\""
+                    .to_string(),
+                "".to_string(),
+                None,
+                None,
+                Vec::new(),
+            ));
+        }
+    };
+    Ok(key)
+}
+
 #[derive(Clone, Debug, Default)]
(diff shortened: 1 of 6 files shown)


src/cli/vector_generate.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    let client =
+        Client::with_config(async_openai::config::OpenAIConfig::default().with_api_key(key));
+
+    let mut vector_count = 0;
+    let start = SystemTime::now();
+    for (i, batch) in batches.iter().enumerate() {
+        let batch_start = SystemTime::now();
+        print!("\rEmbedding batch {:?}/{} ", i + 1, batches.len());
```
chvck, line 221: Should this maybe be info! or debug!

Westwooo: Yeah should have been an info!, have made the change.

src/cli/vector_generate.rs
```diff
@@ -0,0 +1,300 @@
...
+                            format!("failed to execute request: {}", e),
+                            "".to_string(),
+                            None,
+                            None,
+                            Vec::new(),
+                        ))
+                    }
+                },
+                () = ctrl_c_fut =>
+                     Err(ShellError::GenericError(
```
chvck, line 242: Is there a way to cancel the call via openai if ctrlc is pressed?

chvck (approved):
I think we probably want some iteration but this is a good point to build from.

src/cli/vector_generate.rs
Westwooo: Not that is immediately obvious, I will make a note of this as an improvement but don't want to block this merge for that.

src/cli/util.rs
```diff
@@ -354,6 +354,34 @@ pub fn duration_to_golang_string(duration: Duration) -> String {
     golang_string
 }
 
+pub fn read_openai_api_key(engine_state: &EngineState) -> Result<String, ShellError> {
```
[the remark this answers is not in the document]
Westwooo: Will do once this is merged

PR #270, opened 2024-02-19
Add support for vector search

Westwooo:


src/cli/search.rs
```diff
@@ -33,15 +33,27 @@ impl Command for Search {
     fn signature(&self) -> Signature {
         Signature::build("search")
             .required("index", SyntaxShape::String, "the index name")
-            .required(
+            .optional(
```
chvck, line 36: If vector is a named param then should query be the same?

src/cli/search.rs
```diff
@@ -33,15 +33,27 @@ impl Command for Search {
...
                 "query",
                 SyntaxShape::String,
                 "the text to query for using a query string query",
             )
             .named(
-                "databases",
```
chvck, line 42: Why has databases been removed?

src/client/http_client.rs
```diff
@@ -790,8 +793,21 @@ impl SearchQueryRequest {
 
     pub fn payload(&self) -> Option<Vec<u8>> {
         match self {
-            Self::Execute { query, timeout, .. } => {
-                let json = json!({ "query": { "query": query }, "ctl": { "timeout": timeout }});
+            Self::Execute {
+                query,
+                timeout,
+                vector,
```
chvck, line 799: I also feel like a new VectorQueryRequest could be nice just to make intent clear and better code clarity, so that we can simplify the if below..

chvck (changes requested):

src/cli/search.rs
Westwooo: I mainly left it as unnamed for the case where we are just doing a text query. Having to do
```
search indexName --query stringQuery
```
Feels a but redundant to me

chvck: In SDK land the query is optional, and we send a match-none query if the user doesn't specify one which is something that maybe we could do and leads into the other comment I'm about to make.

src/cli/search.rs
```diff
@@ -33,15 +33,27 @@ impl Command for Search {
...
+                "vector",
                 SyntaxShape::String,
-                "the databases which should be contacted",
+                "search vector to be queried for",
+                None,
+            )
+            .named(
```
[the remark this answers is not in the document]
chvck: Possibly we should make it consistent with the new commands that you're adding for vector and actually do `vector search`? I think that the query string is technically optional for the search service, it's just that cbshell only does match queries (I think). If we make --query an option then we can just automatically send a match-none when not specified.

Could be worth a chat with @ingenthr to get his opinion.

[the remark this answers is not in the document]
Westwooo: That sounds reasonable with me ti be honest.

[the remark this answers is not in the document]
Westwooo: @chvck I added a new command, let me know what you think

chvck (approved):
I like it.

PR #275, opened 2024-03-06
Improved vector search ux

Westwooo:
Needs a rebase once vector generate is merged, just needed the command on this branch for testing.

Currently vector search takes the vector with which to perform the query as a required positional argument. This PR changes that, so that we just pipe the vector directly into `vector search`. The command takes input as a pure vector, or the vector doc that is output from `vector enrich-text`:

```
> query 'SELECT contentVector FROM `travel-sample` WHERE meta().id = "landmark_10019"' | select contentVector | vector search vectorSearch contentVector --bucket travelSample

> doc get landmark_10019 | flatten | select contentVector | vector searchvectorSearch contentVector --bucket travelSample

> "embed this text for me please" | vector enrich-text | vector search vectorSearch contentVector --bucket travelSample
```
If the bucket or scope are not specified using the flags then they are read from cb-env, else they default to 'default' and '_default' respectively. For context the form of the input streamed into `vector search` in the first two commands is:
```
╭───┬──────────────────╮
│ # │  contentVector   │
├───┼──────────────────┤
│ 0 │ [list 128 items] │
╰───┴──────────────────╯
```
Whereas for the third command it is: 
```
╭───┬───────────────┬───────────────────╮
│ # │      id       │      content      │
├───┼───────────────┼───────────────────┤
│ 0 │ vector-7c3bb4 │ {record 2 fields} │
╰───┴───────────────┴───────────────────╯
```

### DIFF
--- a/src/cli/vector_search.rs
+++ b/src/cli/vector_search.rs
@@ -1,4 +1,5 @@
 use crate::cli::error::{client_error_to_shell_error, unexpected_status_code_error};
+use crate::cli::util::namespace_from_args;
 use crate::cli::util::{cluster_identifiers_from, get_active_cluster, NuValueMap};
 use crate::client::VectorSearchQueryRequest;
 use crate::state::State;
@@ -33,12 +34,12 @@ impl Command for VectorSearch {
 
     fn signature(&self) -> Signature {
         Signature::build("vector search")
-            .required("index", SyntaxShape::String, "the index name")
-            .required(
+            .optional(
                 "vector",
-                SyntaxShape::Any,
-                "search vector to be queried for",
+                SyntaxShape::List(Box::new(SyntaxShape::Decimal)),
+                "the vector used for searching",
             )
+            .required("index", SyntaxShape::String, "the index name")
             .required(
                 "field",
                 SyntaxShape::String,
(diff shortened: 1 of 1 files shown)


src/cli/vector_search.rs
```diff
@@ -34,11 +34,6 @@ impl Command for VectorSearch {
     fn signature(&self) -> Signature {
         Signature::build("vector search")
             .required("index", SyntaxShape::String, "the index name")
-            .required(
```
chvck, line 37: I suspect that this should still be an input field. See `doc get` where it supports piping but is also an optional field.

chvck (approved):

src/cli/vector_search.rs
```diff
@@ -85,19 +93,88 @@ fn run(
...
+                        "Please supply vector or output from `vector enrich-text`".to_string(),
+                        "".to_string(),
+                        None,
+                        None,
+                        vec![e],
+                    ));
+                }
+            };
+
+            if rec.0 == vec!["id", "content"] {
```
chvck, line 117: I'm slightly suspicious of this

Westwooo: I am also, will have a think about how to improve

PR #277, opened 2024-03-08
Add sub-doc get command

Westwooo:


chvck (commented):
We should discuss this, by subdoc I originally meant like how SDKs do it but this might also be an interesting approach. I do think we probably want to let users fetch multiple fields at once though.

src/cli/sub_doc_get.rs
```diff
@@ -0,0 +1,335 @@
...
+    }
+
+    Ok(Value::List {
+        vals: results,
+        span: call.head,
+    }
+    .into_pipeline_data())
+}
+
+pub(crate) fn ids_from_input(
```
chvck, line 238: Everything from here down is duplicated code I think?

chvck (commented):
Following conversation last week I think this approach is probably correct, just 1 comment.

src/client/kv_client.rs
```diff
@@ -284,13 +299,40 @@ impl KvClient {
...
                 let content = if let Some(body) = r.0.body() {
-                    match serde_json::from_slice(body.as_ref()) {
-                        Ok(v) => Some(v),
-                        Err(e) => {
-                            return Err(ClientError::RequestFailed {
-                                reason: Some(e.to_string()),
-                                key: r.1,
-                            });
+                    // Leading 0 indicates response to subdoc multi lookup
+                    if body[0] == 0 {
```
chvck, line 303: Can we do this off op_code from the result?

src/cli/subdoc_get.rs
```diff
@@ -0,0 +1,530 @@
...
+            Example{
+                description: "Fetches address and content fields from multiple documents with IDs from the previous command",
+                example: "[landmark_10019, landmark_10020] | subdoc get [address, content]",
+                result: None
+            },
+        ]
+    }
+}
+
+fn run_subdoc_get(
```
chvck, line 135: Is there a way we can DRY these 2 calls up a bit? They're almost identical aside from the op types.

src/client/kv_client.rs
Westwooo: Done

src/cli/subdoc_get.rs
Westwooo: Done

src/cli/subdoc_get.rs
```diff
@@ -0,0 +1,382 @@
...
+        for ids in all_ids.clone() {
+            for id in ids {
+                let deadline = Instant::now().add(active_cluster.timeouts().data_timeout());
+
+                let ctrl_c = ctrl_c.clone();
+                let id = id.clone();
+
+                let client = client.clone();
+
+                let copy = paths.to_vec();
```
chvck, line 203: Can this just be paths.clone()?

src/cli/subdoc_get.rs
```diff
@@ -0,0 +1,382 @@
...
+            self.key.unwrap_or_default(),
+            span,
+        );
+        collected.add("content", self.content.unwrap_or_default());
+        collected.add_i64("cas", self.cas.unwrap_or_default(), span);
+        collected.add_string("error", self.error.unwrap_or_default(), span);
+        collected.add_string("cluster", self.cluster, span);
+        collected.into_value(span)
+    }
+}
```
chvck, line 382: I think all this already exists in doc_get?

src/cli/subdoc_get.rs
```diff
@@ -0,0 +1,382 @@
...
+                    key: id.clone(),
+                    path: paths[0].clone(),
+                };
+
+                if paths.len() > 1 {
+                    request = KeyValueRequest::SubdocMultiLookup {
+                        key: id,
+                        paths: copy,
+                    };
+                }
```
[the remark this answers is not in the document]
Westwooo: Done

src/cli/subdoc_get.rs
Westwooo: It is, have removed

src/cli/doc_get.rs
```diff
@@ -274,7 +274,7 @@ pub(crate) fn ids_from_input(
 }
 
 #[derive(Debug)]
-struct GetResult {
+pub struct GetResult {
```
chvck, line 277: Nit: Should probably be pub(crate) but doesn't matter enough to block merging.

Westwooo: Will change, then merge after building

chvck (approved):

PR #278, opened 2024-03-19
Initial implementation of vector enrich-doc

Westwooo:


src/cli/vector_enrich_text.rs
```diff
@@ -167,7 +167,7 @@ fn vector_enrich_text(
     let start = SystemTime::now();
     for (i, batch) in batches.iter().enumerate() {
         let batch_start = SystemTime::now();
-        info!("Embedding batch {:?}/{} ", i + 1, batches.len());
+        println!("Embedding batch {:?}/{} ", i + 1, batches.len());
```
Westwooo, line 170: I changed to println!() since these commands can take a little time I wanted the User to have an indicator of progress without running in info mode, since this log level is quite verbose.

src/cli/llm_client.rs
```diff
@@ -53,15 +54,24 @@ impl OpenAIClient {
 
         debug!("Total tokens: {:?}\n", tokens.len());
 
+        //Regardless of token limit OpenAI's API can only accept batches up to 2048 in length
         let num_batches = (tokens.len() / self.max_tokens) + 1;
-        let batch_size = chunks.len() / num_batches;
+        let batch_size = cmp::min(2047, chunks.len() / num_batches);
```
chvck, line 59: We should log when we coerce this.

src/cli/vector_enrich_doc.rs
```diff
@@ -0,0 +1,352 @@
...
+        input: PipelineData,
+    ) -> Result<PipelineData, ShellError> {
+        vector_enrich_doc(self.state.clone(), engine_state, stack, call, input)
+    }
+
+    fn examples(&self) -> Vec<Example> {
+        vec![
+            Example {
+                description: "Open local json doc and enrich the field named 'description'",
+                example: "open ./local.json | vector enrich-doc",
```
chvck, line 88: Why doesn't think require a field param?

Westwooo: It does, I forgot it

src/cli/llm_client.rs
chvck: Maybe should also be println?

PR #281, opened 2024-03-22
Read api_key for llm from config file

Westwooo:


src/cli/util.rs
```diff
@@ -354,24 +354,13 @@ pub fn duration_to_golang_string(duration: Duration) -> String {
     golang_string
 }
 
-pub fn read_openai_api_key(engine_state: &EngineState) -> Result<String, ShellError> {
-    let key = match engine_state.get_env_var("OPENAI_API_KEY") {
```
chvck, line 358: I think it's probably also worth being able to read from env var.

chvck (approved):

PR #282, opened 2024-03-22
Ask command

Westwooo:


src/cli/llm_client.rs
```diff
@@ -113,4 +128,62 @@ impl OpenAIClient {
 
         Ok(rec)
     }
+
+    async fn ask(&self, question: String, context: Vec<String>) -> Result<String, ShellError> {
+        let mut messages: Vec<ChatCompletionRequestMessage> = vec![];
+
+        messages.push(
+            ChatCompletionRequestSystemMessageArgs::default()
+                .content("You are a helpful assistant.")
```
chvck, line 137: Going to need a comment on some context around why we need this :)

src/cli/llm_client.rs
```diff
@@ -113,4 +128,62 @@ impl OpenAIClient {
...
+            .max_tokens(512u16)
+            .model("gpt-3.5-turbo")
+            .messages(messages)
+            .build()
+            .unwrap();
+
+        let response = client.chat().create(request).await;
+
+        let answer = match response {
+            Ok(r) => r.choices[0].message.content.as_ref().unwrap().to_string(),
```
chvck, line 175: What happens if the response has no choices? Is that possible?

src/cli/ask.rs
```diff
@@ -0,0 +1,124 @@
...
+    }
+}
+
+impl Command for Ask {
+    fn name(&self) -> &str {
+        "ask"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("ask")
```
chvck, line 34: I wonder if this is the right command name for this? Not that I have anything better...

Westwooo: I'm a fan of ask, seems fairly clear what it does and has the advantage of being short

src/cli/llm_client.rs
Westwooo: Done

src/cli/llm_client.rs
Westwooo: I don't think this is possible. You get a response for each message sent, and we are sending one, so will always be a choice. If there is no message, which there can't be, or another issue with the request then we end up in the Err() branch.

src/cli/ask.rs
```diff
@@ -0,0 +1,131 @@
...
+    engine_state: &EngineState,
+    stack: &mut Stack,
+    call: &Call,
+    input: PipelineData,
+) -> Result<PipelineData, ShellError> {
+    let span = call.head;
+
+    let question: String = call.req(engine_state, stack, 0)?;
+    let mut context: Vec<String> = Vec::new();
+    match input.into_value(span) {
```
chvck, line 70: I'm still a little dubious about us doing this everywhere, unless nushell also do it? Although it does work so...

src/cli/ask.rs
```diff
@@ -0,0 +1,131 @@
...
+        Value::List {
+            vals,
+            internal_span: span,
+        } => {
+            for v in vals {
+                let rec = match v.as_record() {
+                    Ok(r) => r,
+                    Err(e) => {
+                        return Err(ShellError::GenericError {
+                            error: "Supply a table of strings".to_string(),
```
chvck, line 80: Probably best to use some sort of JSON terminology rather than nushell

src/cli/ask.rs
Westwooo: I agree, does feel a little hacky, and I can't see that nu-shell do it. But as you say it does work, and I'm not sure how else to achieve the same effect otherwise

src/cli/ask.rs
Westwooo: In principle I agree but in  practice I think it makes this command more confusing. The context piped in needs to be a single column of a table, which is the equivalent to a list of json objects with a single field. I think asking for a single table column, with an example in the --help is clearer than asking for a list of json objects that contain a single field.

chvck (approved):

PR #287, opened 2024-03-25
Take embedding source text as an optional positional arg

Westwooo:


src/cli/vector_enrich_text.rs
```diff
@@ -276,6 +215,99 @@ fn vector_enrich_text(
...
+                            "Could not parse list of files".to_string(),
+                            "".to_string(),
+                            None,
+                            None,
+                            vec![e],
+                        ));
+                    }
+                };
+
+                let index = match rec.0.iter().position(|r: &String| *r == "name") {
```
chvck, line 248: I feel like this should be a name_column or something akin to id_column in the doc commands.

chvck (approved):

PR #293, opened 2024-04-10
Adds support for named and positional query parameters

Westwooo:


src/cli/query.rs
```diff
@@ -61,6 +62,12 @@ impl Command for Query {
                 "the scope to query against",
                 None,
             )
+            .named(
+                "params",
```
chvck, line 66: query-params perhaps? I don't know, just feel like using params might come back to bite us in the future?

src/cli/query.rs
```diff
@@ -95,6 +102,24 @@ fn query(
...
     let statement: String = call.req(engine_state, stack, 0)?;
 
+    let params: Option<serde_json::Value> =
+        match call.get_flag::<Value>(engine_state, stack, "params")? {
+            Some(p) => match p {
+                Value::Record { .. } => Some(convert_nu_value_to_json_value(&p, span).unwrap()),
+                Value::List { .. } => Some(convert_nu_value_to_json_value(&p, span).unwrap()),
+                _ => {
+                    return Err(ShellError::GenericError {
+                        error: "Parameters must be a list or record".to_string(),
```
chvck, line 112: Maybe we should use object rather than record, users are unlikely to know the nu lingo.

src/client/http_client.rs
```diff
@@ -664,23 +665,48 @@ impl QueryRequest {
                 scope,
                 timeout,
                 transaction,
+                parameters,
             } => {
-                let mut json = HashMap::new();
+                let mut json: HashMap<String, serde_json::Value> = HashMap::new();
```
chvck, line 670: Why did we have to change this to be annotated?

src/client/http_client.rs
```diff
@@ -664,23 +665,48 @@ impl QueryRequest {
...
+                }
+
+                if let Some(params) = parameters {
+                    match params {
+                        serde_json::Value::Array(_) => {
+                            json.insert("args".to_string(), params.clone());
+                        }
+                        serde_json::Value::Object(map) => {
+                            for (k, v) in map.iter() {
+                                json.insert(format!("${}", *k), v.clone());
```
chvck, line 703: We're going to need to add a check here to see if the key already has a $ prefix, if so then don't add it ourselves.

chvck: I also think we should just convert this to &str rather than .to_string() everywhere else.

src/cli/query.rs
Westwooo: In the docs they are called place-holder parameters, but that feels too long winded. Similarly query-params seems a bit lengthy. What is your concern with params?

src/cli/query.rs
Westwooo: Good point, will change this now

src/client/http_client.rs
Westwooo: We don't I think I just did it while getting things working, will change now

src/client/http_client.rs
```diff
@@ -664,23 +665,53 @@ impl QueryRequest {
...
+                    match params {
+                        serde_json::Value::Array(_) => {
+                            json.insert("args".to_string(), params.clone());
+                        }
+                        serde_json::Value::Object(map) => {
+                            for (k, v) in map.iter() {
+                                let key = if k.starts_with("$") {
+                                    k.clone()
+                                } else {
+                                    format!("${}", *k)
```
chvck, line 706: I still think we should convert this to &str rather than everything else to string. Do you disagree?

Westwooo: I think that we need to have the HashMap be <String, serde_json::Value> so that we can correctly format the positional parameters above. If we have it as a <String, string> then the payload ends up with:
```
args: "{"LAX", 13000}",  - the whole thing is a string
```
But with the <String, Value> map we get what we need where the array is printed as an array.

Westwooo: I don't think the compiler will let me. We get `borrowed value does not live long enough` since key leaves scope after the if, but we still use the reference in the json, so I think it needs to be a String.

chvck (approved):

PR #305, opened 2024-04-18
Unset cbenv scope/col when higher entity changed

Westwooo:
Currently when a bucket/scope is changed using cb-env the scope/collection remain unchanged. We should change this so that when a bucket is changed the scope & collection are unset, unless specified using the --preserve flag.


src/remote_cluster.rs
```diff
@@ -127,6 +127,11 @@ impl RemoteCluster {
         *active = Some(name);
     }
 
+    pub fn unset_active_scope(&self) {
```
chvck, line 130: Should these new functions just be parameters to set_active_bucket and set_active_scope?

Westwooo: Done!

chvck (approved):

PR #310, opened 2024-04-30
Improve testing framework

Westwooo:


src/cli/tests/utils.rs
```diff
@@ -0,0 +1,76 @@
...
+        }
+        ClusterType::Mock => ClusterUnderTest::Mocked(MockCluster::start(loaded_config).await),
+    };
+    println!("Cluster: {:?}", &server);
+
+    Arc::new(server)
+}
+
+pub fn test_cluster() -> Arc<ClusterUnderTest> {
+    println!("JW IS HERE 1");
```
chvck, line 33: :)

tests/doc_get.rs
```diff
@@ -1,15 +1,17 @@
...
+fn test_get_a_document() {
+    let cluster = utils::test_cluster();
+
     if !cluster.config().supports_feature(TestFeature::KeyValue) {
-        return true;
+        return;
     }
 
-    playground::CBPlayground::setup("get_a_document", cluster.config(), None, |dirs, sandbox| {
+    CBPlayground::setup("get_a_document", cluster.config(), None, |dirs, sandbox| {
```
chvck, line 14: Given we're going to use a global variable effectively now it'd be cool if CBPlayground::setup also just returned the cluster to us rather than needing us to supply it.  We can just tell it which feature(s) we're testing as a part of the parameters we pass in. The tricky part is how to signal from setup that the test should be skipped.

chvck:
It might be worth having a google to see how others have approach the "setup suite" issue. When I looked it seems like there are crates out there that can help a bit.

chvck:
FWIW I think integration tests should also remain in the tests directory, and unit tests alongside the code. Another thing to consider is whether we can add some unit tests for trickier cases to repro, see the mocking approach in https://github.com/couchbaselabs/couchbase-rs/pull/155. Out of scope for this PR though.

Westwooo:
> FWIW I think integration tests should also remain in the tests directory, and unit tests alongside the code. Another thing to consider is whether we can add some unit tests for trickier cases to repro, see the mocking approach in [couchbaselabs/couchbase-rs#155](https://github.com/couchbaselabs/couchbase-rs/pull/155). Out of scope for this PR though.

I agree, have moved everything back.

tests/doc_get.rs
Westwooo: Done an initial iteration of this. One issue is that we call setup when starting the standalone cluster when we wait for kv/scope/collection. This is why I have added an Option to pass in the config, since in those cases we need to pass the parsed config in, otherwise you get stuck in a infinite loop, but there is probably a cleaner way to handle it. 

If a test feature is not supported then setup just returns, which will mark the tests as passed, not ignored. This is the same behaviour as before the change, so no worse off. Unfortunately I couldn't find a way to ignore tests at runtime, but I'll take another stab at it tomorrow.

Cargo.toml
```diff
@@ -83,20 +83,26 @@ nu-test-support = { version = "0.83.1"}
...
 strum_macros = "0.24.3"
 
-
 [features]
 # Enable to statically link OpenSSL; otherwise the system version will be used. Not enabled by default because it takes a while to build
 static-link-openssl = ["dep:openssl"]
 
+# We use a list of supported features to determine which tests we want to run, the defaults will always be run, then we
+# define any additional features using the specific server versions we test against
+default = ["key_value", "query", "query_index", "query_index_definitions", "query_index_advise"]
```
chvck, line 92: Do we have a way to *only* run tests against a specific feature like query?

tests/common/playground.rs
```diff
@@ -184,15 +192,15 @@ default-collection = \"{}\"
...
-                RetryExpectations::ExpectNoOut => {
-                    if out.out.is_empty() {
-                        return;
-                    } else {
-                        println!("Expected no out but was {}", out.out);
-                        sleep(interval);
-                        continue;
-                    }
-                }
+                // RetryExpectations::ExpectNoOut => {
```
chvck, line 195: Why commented out?

.github/workflows/ci.yml
```diff
@@ -86,12 +86,13 @@ jobs:
     - name: Test
       uses: actions-rs/cargo@v1
       env:
+        CONN_STRING: https://127.0.0.1:8091
```
chvck, line 89: If this is only running unit tests then why does it need CONN_STRING?

Cargo.toml
Westwooo: Yeah can do: `cargo test --no-default-features --features query`

tests/common/playground.rs
Westwooo: It wasn't being used any where on main, but now that I've moved it, the compiler is picking it up as dead code. But I thought we might use it in the future, so I commented it out instead of removing.

chvck: I'm a bit confused why this code isn't used anymore.

tests/query_indexes.rs
```diff
@@ -175,28 +98,28 @@ pub async fn test_should_get_indexes_with_context(cluster: Arc<ClusterUnderTest>
...
             let fields = "`field1`,`field2`";
             let index_name1 =
-                create_index(cmd.clone(), fields, keyspace.clone(), dirs.test(), sandbox);
+                utils::create_index(cmd.clone(), fields, keyspace.clone(), dirs.test(), sandbox);
             let index_name2 =
-                create_index(cmd.clone(), fields, keyspace.clone(), dirs.test(), sandbox);
+                utils::create_index(cmd.clone(), fields, keyspace.clone(), dirs.test(), sandbox);
+
+            // Let the indexes be created
+            sleep(time::Duration::from_secs(20));
```
chvck, line 122: We should use a consistency util rather than a sleep.

tests/common/playground.rs
Westwooo: I also am, but I don't think that it's gone from being used to not. It never appears to have been used but I think the re-arrange of where the tests are now means that it is flagged as dead code.

.github/workflows/ci.yml
Westwooo: An artefact from trying to get things passing, but yes we don't need it.

tests/common/playground.rs
```diff
@@ -20,6 +20,7 @@ pub struct PerTestOptions {
     no_default_collection: bool,
 }
 
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
Westwooo: I think it's because the method `set_no_default_collection` is only used in tests, so the compiler thinks that it isn't used. We could get around this by creating a test mod and putting this in in but that gets a bit messier so it's easier to just allow it imo.

tests/query_indexes.rs
Westwooo: Done!

tests/common/playground.rs
```diff
@@ -0,0 +1,449 @@
...
+                    println!("Retry func returned fail")
+                }
+                Err(e) => println!("Retry func returned error: {}", e),
+            }
+
+            sleep(interval);
+        }
+    }
+
+    pub async fn start(c: Config) -> Arc<TestConfig> {
```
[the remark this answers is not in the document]
Westwooo: Done!

PR #313, opened 2024-05-17
Add gemini client

Westwooo:


src/cli/ask.rs
```diff
@@ -82,46 +80,51 @@ pub fn ask(
...
-                if rec.columns().len() > 1 {
-                    return Err(ShellError::GenericError {
-                        error: "Use 'select' to choose a single column to pipe to 'ask'"
-                            .to_string(),
-                        msg: "".to_string(),
-                        span: Some(span),
-                        help: None,
-                        inner: vec![],
-                    });
+                    let ctx_string = rec.get_index(0).unwrap().1.as_str()?;
```
chvck, line 116: Is it possible for rec to be `{}`?

src/client/gemini_client.rs
```diff
@@ -0,0 +1,287 @@
...
+// According to the Gemini API docs: A token is equivalent to about 4 characters for Gemini models
+const CHARS_PER_TOKEN: usize = 4;
+
+const MAX_EMBEDDING_DIMENSION: u32 = 768;
+
+// At most 100 requests can be in one batch
+const MAX_BATCH_SIZE: usize = 100;
+
+impl GeminiClient {
+    pub fn new(api_key: String, max_tokens: Option<usize>) -> Self {
```
chvck, line 23: This just like a "this is a nice feature we could use" in future, rather than a suggestion to change:
`max_tokens: impl Into<Option<usize>>`
then `let max_tokens = max_tokens.into()`. It lets use called the function with max_tokens just set to something like 5000 rather than Some(5000)

src/client/gemini_client.rs
```diff
@@ -0,0 +1,287 @@
...
+                inner: vec![],
+            });
+        }
+
+        let url = format!(
+            "https://generativelanguage.googleapis.com/v1/models/text-embedding-004:batchEmbedContents?key={}",
+            self.api_key
+        );
+
+        let mut batch_request = EmbeddingBatchRequest { requests: vec![] };
```
chvck, line 74: I think this would be nicer as EmbeddingBatchRequest::new() where it just automatically sets requests to vec![] inside of new?

src/client/gemini_client.rs
```diff
@@ -0,0 +1,287 @@
...
+            span: None,
+            help: None,
+            inner: vec![],
+        });
+    };
+
+    Ok(bytes)
+}
+
+async fn execute_request<T>(url: String, json_body: T) -> Result<Response, ShellError>
```
chvck, line 201: Slight concerns about this not being deadlined.

Westwooo: Have added a 30 second timeout

src/cli/ask.rs
Westwooo: It could be, have added handling of that case

src/client/gemini_client.rs
Westwooo: Done!

src/client/gemini_client.rs
Westwooo: That's a nice idea, have made the change.

src/client/gemini_client.rs
```diff
@@ -0,0 +1,308 @@
...
+const CHARS_PER_TOKEN: usize = 4;
+
+const MAX_EMBEDDING_DIMENSION: u32 = 768;
+
+// At most 100 requests can be in one batch
+const MAX_BATCH_SIZE: usize = 100;
+
+impl GeminiClient {
+    pub fn new(api_key: String, max_tokens: impl Into<Option<usize>>) -> Self {
+        let max_tokens = match max_tokens.into() {
```
chvck, line 25: Another not saying we should do I think we could actually something like `let max_tokens = max_tokens.into().unwrap_or(MAX_FREE_TIER_TOKENS)`.

src/client/gemini_client.rs
chvck: This seems like it should be customizable or at least passed in from somewhere higher? Fine with this filed as an issue to address later though.

chvck (approved):

PR #318, opened 2024-05-29
Add path not found error for subdoc get

Westwooo:
Before this change, a missing path would result in the following: 
```
> subdoc get names landmark_10019
╭───┬────────────────┬─────────┬─────┬────────────────┬─────────╮
│ # │       id       │ content │ cas │     error      │ cluster │
├───┼────────────────┼─────────┼─────┼────────────────┼─────────┤
│ 0 │ landmark_10019 │         │   0 │ Request failed │ local   │
╰───┴────────────────┴─────────┴─────┴────────────────┴─────────╯
```

Now we get: 
```
> subdoc get names landmark_10019
╭───┬────────────────┬─────────┬─────┬────────────────┬─────────╮
│ # │       id       │ content │ cas │     error      │ cluster │
├───┼────────────────┼─────────┼─────┼────────────────┼─────────┤
│ 0 │ landmark_10019 │         │   0 │ Path not found │ local   │
╰───┴────────────────┴─────────┴─────┴────────────────┴─────────╯
```

### DIFF
--- a/src/client/error.rs
+++ b/src/client/error.rs
@@ -84,6 +84,10 @@ pub enum ClientError {
         reason: String,
         address: String,
     },
+    PathNotFound {
+        key: String,
+        path: String,
+    },
 }
 
 impl ClientError {
@@ -95,6 +99,7 @@ impl ClientError {
             ClientError::Timeout { key, .. } => key.clone(),
             ClientError::Cancelled { key } => key.clone(),
             ClientError::RequestFailed { key, .. } => key.clone(),
+            ClientError::PathNotFound { key, .. } => Some(key.clone()),
             _ => None,
         }
     }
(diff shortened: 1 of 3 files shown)


src/client/error.rs
```diff
@@ -191,12 +200,14 @@ impl ClientError {
         reason: Option<String>,
         key: String,
         cid: u32,
+        path: String,
```
chvck, line 203: Should probably be Option?

src/client/kv.rs
```diff
@@ -271,12 +271,15 @@ impl KvEndpoint {
         rx: Receiver<KvResponse>,
         key: String,
         cid: u32,
+        path: String,
```
chvck, line 274: Should be Option I think?

Westwooo: Done!

chvck (approved):

PR #321, opened 2024-05-31
Fix parsing of sample import response

Westwooo:


.github/bin/init-cluster
```diff
@@ -31,7 +31,7 @@ options = {
   password: ENV.fetch("CB_PASSWORD", "password"),
   bucket: ENV.fetch("CB_BUCKET", "default"),
   sec_bucket: ENV.fetch("CB_SEC_BUCKET", ""),
-  server_quota: ENV.fetch("CB_SERVER_QUOTA", 2048).to_i,
```
chvck, line 34: Why this change? We should probably also move to using dinocluster instead of this but that's a different topic...

src/cli/buckets_sample.rs
```diff
@@ -103,32 +86,41 @@ fn load_sample_bucket(
...
+                collected.add_string("status", "success", span);
+            }
             _ => {
-                return Err(unexpected_status_code_error(
-                    response.status(),
-                    response.content(),
-                    span,
-                ))
+                // The format of the returned error varies between server versions so we have to do this substring matching
+                let error = if response.content().contains("already loaded") {
```
chvck, line 106: Typically I'd say that string matching is going to lead to pain down the road. Repeatedly we've seen that the server changes the message and error handling gets broken on clients. Admittedly there's not really a better way but can we just continue surfacing the error message from the server directly?

tests/buckets_sample.rs
```diff
@@ -0,0 +1,56 @@
+mod common;
```
chvck, line 1: Nice!

.github/bin/init-cluster
Westwooo: I need to hit the RAM quota limit to test the error that we get when trying to import a sample bucket with less that 200MB left. But I think I'm going to leave the quota and create some temp bucket to take up the space required.

tests/query.rs
```diff
@@ -139,10 +139,10 @@ fn fetch_meta() {
                 playground::RetryExpectations::ExpectOut,
                 |json| -> TestResult<bool> {
                     val = json.clone();
-                    Ok(true)
+                    // Wait until the document has been indexed
+                    Ok(val["results"].as_array().unwrap().len() == 1)
```
Westwooo, line 143: This test was flaky due to a race condition where the index would be created but the document created in the test had not been indexed yet, so the query would succeed but there would be no results.

src/cli/buckets_sample.rs
```diff
@@ -103,32 +86,44 @@ fn load_sample_bucket(
...
+            }
             _ => {
-                return Err(unexpected_status_code_error(
-                    response.status(),
-                    response.content(),
-                    span,
-                ))
+                // Errors are in the form: `["error msg here"]` so need to drop the first and last 2 chars
+                let mut chars = response.content().chars();
+                chars.next();
```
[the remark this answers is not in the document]
Westwooo: We can, done!

src/cli/buckets_sample.rs
```diff
@@ -103,32 +86,39 @@ fn load_sample_bucket(
...
             _ => {
-                return Err(unexpected_status_code_error(
-                    response.status(),
-                    response.content(),
-                    span,
-                ))
+                let msg = if response.content().contains("\"errors\":") {
+                    // In 7.6 the content is in the form '"{"errors":["Err msg here"]}"' so we need to remove the prepended
+                    // "errors:" and the extra trailing char
+                    &response.content()[12..response.content().len() - 2]
```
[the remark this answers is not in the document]
Westwooo: That's why I went with the string matching approach originally, since trying to trim the message from the server just opens a whole can of worms

chvck (approved):

PR #330, opened 2024-06-12
Add model flag

Westwooo:


src/client/bedrock_client.rs
```diff
@@ -29,52 +29,102 @@ impl BedrockClient {
...
+                                },
+                                InvokeModelError::AccessDeniedException(inner_err) => {
+                                    (inner_err.message.as_ref().unwrap().to_string(),
+                                    Some("Have you been granted access to this model in the AWS web console?".to_string()))
+                                },
+                                InvokeModelError::ValidationException(inner_err) => {
+                                    let msg = inner_err.message.as_ref().unwrap().to_string();
+                                    let help = None;
+
+                                    let (msg, help) = if msg.contains("extraneous key [dimensions] is not permitted") {
```
chvck, line 87: I guess there's no error codes we can use? Just aware this is going to break as soon as they change the text.

src/client/gemini_client.rs
```diff
@@ -67,45 +66,54 @@ impl GeminiClient {
         &self,
         batch: &Vec<String>,
         dim: Option<usize>,
+        model: String,
     ) -> Result<Vec<Vec<f32>>, ShellError> {
-        let dimension = match dim {
-            Some(d) => {
-                if d > MAX_EMBEDDING_DIMENSION || d < 1 {
+        if !model.contains("models/") {
```
chvck, line 71: Could we just prefix this automatically if missing and log something at info?

chvck (commented):
Should we maybe put model in the config file and make it an optional override for the commands? That way the user doesn't have to specify it every time if they just use the same model.

src/client/bedrock_client.rs
Westwooo: None that I could find unfortunately, I know it fragile but I think this is the best we can do.

src/client/gemini_client.rs
Westwooo: Done!

src/client/bedrock_client.rs
chvck: Are the messages the provider give us actually fairly descriptive anyway? Thinking that users might have more lucky googling those error messages if they need to.

Westwooo: Yeah you're right, this is the result of not doing the string matching: 
```
> vector enrich-text testing --model another-llm-model
Embedding batch 1/1
Error:   × The provided model identifier is invalid.
  help: Check that the model name is correct and that it supports the chosen dimensionality.

> vector enrich-text testing --dimension 1000000 --model amazon.titan-embed-text-v2
Embedding batch 1/1
Error:   × The provided model identifier is invalid.
  help: Check that the model name is correct and that it supports the chosen dimensionality.
  
  > vector enrich-text testing --dimension 1000000 --model amazon.titan-embed-text-v2:0
Embedding batch 1/1
Error:   × Malformed input request: #: only 1 subschema matches out of 2, please reformat your input and try again.
  help: Check that the model name is correct and that it supports the chosen dimensionality.
```

chvck (approved):

PR #332, opened 2024-06-24
Initial support for Capella testing

Westwooo:
Can now run tests locally against Capella cluster, and we have a feature that ignores tests that we do not expect to pass. Not yet adding github workflow since we don't really want to be spinning up Capella clusters on every PR. Will speak to Sriram about handing running of Capella tests over to QE.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,4 @@
+workspace = { members = ["utilities"] }
 [package]
 name = "couchbase-shell"
 version = "0.95.0"
@@ -55,8 +56,8 @@ tokio = { version = "1.38.0", features = ["full"] }
 tokio-util = { version = "0.7.11", features = ["codec"] }
 tokio-rustls = { version = "0.26.0" }
 toml = "0.8.14"
-trust-dns-resolver = { version = "0.23.2", features = ["dns-over-rustls"] }
 uuid = { version = "1.9.1", features = ["v4"] }
+utilities = { path = "utilities" }
 webpki-roots = "0.26.3"
 
 [target.'cfg(not(any(target_os = "macos", target_os = "windows")))'.dependencies]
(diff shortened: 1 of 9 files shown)


tests/common/playground.rs
```diff
@@ -447,3 +520,47 @@ pub enum RetryExpectations {
     //ExpectNoOut
     AllowAny { allow_out: bool, allow_err: bool },
 }
+
+fn try_lookup_srv(addr: String) -> Vec<String> {
```
Westwooo, line 524: This is almost the same as the `try_lookup_srv` in the Client, except this doesn't return a client error. I think the only way to have the function shared would be to create an external utilities crate, since this would be one of the only functions in there I'm not sure that it's worth the trouble.

tests/common/playground.rs
```diff
@@ -382,23 +391,55 @@ impl CBPlayground {
...
         let mut params = HashMap::new();
         params.insert("name", scope_name.clone());
 
-        let client = reqwest::Client::new();
-        let res = client
-            .post(uri)
-            .form(&params)
-            .basic_auth(username, Some(password))
-            .send()
+        let mut client_builder = ClientBuilder::new();
```
chvck, line 401: Seems like we should probably have a utility function for building the http client?

tests/common/playground.rs
```diff
@@ -413,26 +426,28 @@ impl CBPlayground {
...
-        let res = client
-            .post(uri)
-            .form(&params)
-            .basic_auth(username, Some(password))
-            .send()
-            .await
-            .unwrap();
+        let uri = build_uri(conn_string.clone(), bucket, scope).await;
+
+        let res = task::spawn_blocking(move || {
```
chvck, line 438: I'm pretty sure that reqwest has a blocking API doesn't it?

Westwooo: Resolving as have discussed on slack

Cargo.toml
```diff
@@ -83,14 +84,16 @@ static-link-openssl = ["dep:openssl"]
 # We use a list of supported features to determine which tests we want to run, the defaults will always be run, then we
 # define any additional features using the specific server versions we test against
 default = ["key_value", "query", "query_index", "query_index_definitions", "query_index_advise"]
-"7.1.1" = ["collections"]
-"7.0.3" = ["collections"]
+"7.1.1" = ["collections", "import_sample"]
+"7.0.3" = ["collections", "import_sample"]
+"Capella" = ["collections"]
```
chvck, line 89: Really pick but I think features are generally lower case first letter?

Westwooo: I did not know that, done!

tests/common/playground.rs
```diff
@@ -377,28 +386,33 @@ impl CBPlayground {
...
-            .post(uri)
-            .form(&params)
-            .basic_auth(username, Some(password))
-            .send()
-            .await
-            .unwrap();
+        let uri = build_uri(conn_string.clone(), bucket, None).await;
+
+        // Send creates it's own runtime so need to spwn a thread where blocking is allowed to do this in an async fn
+        let res = task::spawn_blocking(move || {
```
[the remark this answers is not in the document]
Westwooo: I think this was done when trying to get the code running, have changed back to async client.

chvck (approved):

PR #341, opened 2024-07-09
Initial implementation of v4 Capella cluster management

Westwooo:


src/state.rs
```diff
@@ -240,42 +247,11 @@ impl State {
...
-                })
-            }
-        };
-        org.set_id(id);
-        Ok(())
-    }
-
     pub fn capella_org_for_cluster(
-        &self,
+        &mut self,
```
chvck, line 251: Same as below, why mutable references?

src/state.rs
```diff
@@ -292,6 +268,39 @@ impl State {
         }
     }
 
+    pub fn active_project(&mut self) -> Result<&mut RemoteCapellaProject, ShellError> {
```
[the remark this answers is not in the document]
Westwooo: Because when we do the fetching of the project ID we set the id field of the `RemoteCapellaProject` so it needs to be mutable. The same applies for the organization.

chvck (approved):

src/state.rs
```diff
@@ -232,56 +236,28 @@ impl State {
...
-            Some(a) => a,
-            None => {
-                return Err(ShellError::GenericError {
-                    error: "No active Capella organization set".to_string(),
-                    msg: "".to_string(),
-                    span: None,
-                    help: None,
-                    inner: Vec::new(),
-                })
+        if let Some(org) = self.capella_orgs.get(&active) {
```
chvck, line 239: We should probably combine this with the above check as a single match statement or similar.

Westwooo: Done

chvck (approved):

PR #346, opened 2024-07-17
V4 buckets

Westwooo:


src/client/cloud.rs
```diff
@@ -17,7 +17,7 @@ use tokio::runtime::Runtime;
 use tokio::{select, time::Instant};
 
 const CLOUD_URL: &str = "https://cloudapi.cloud.couchbase.com";
-pub const CAPELLA_SRV_SUFFIX: &str = "cloud.couchbase.com";
+pub const CAPELLA_CONNSTR_PREFIX: &str = "couchbases://";
```
Westwooo, line 20: Change to `couchbases` prefix from the suffix otherwise this does not work with local dev

src/cli/buckets_config.rs
```diff
@@ -99,3 +114,66 @@ fn buckets(
...
+        &client,
+        deadline,
+        span,
+        org_id.clone(),
+        project_id.clone(),
+    )?;
+
+    client
+        .capella_request(
+            CapellaRequest::GetBucketV4 {
```
chvck, line 169: I don't think this returns the bucket config? This command was intended to be quite low level and return the config as reported by /pools/default/b/<bucket>

src/remote_cluster.rs
```diff
@@ -14,7 +14,7 @@ pub enum RemoteClusterType {
 
 impl From<Vec<String>> for RemoteClusterType {
     fn from(hostnames: Vec<String>) -> Self {
-        if hostnames.len() == 1 && hostnames[0].contains(&CAPELLA_SRV_SUFFIX.to_string()) {
+        if hostnames.len() == 1 && hostnames[0].contains(&CAPELLA_CONNSTR_PREFIX.to_string()) {
```
chvck, line 17: This means that any cluster registered with a single address and `couchbases://` will be registered as Provisioned? That doesn't seem correct?

src/cli/buckets_config.rs
Westwooo: For some reason I thought that this endpoint wouldn't work for Capella clusters, so was just doing a get that printed out all the fields we get. But it does work so have changed it back

src/remote_cluster.rs
```diff
@@ -3,18 +3,19 @@ use crate::{
...
+#[derive(Debug, Deserialize, Serialize, Clone, Copy, Eq, PartialEq)]
 pub enum RemoteClusterType {
     Provisioned,
     Other,
 }
 
 impl From<Vec<String>> for RemoteClusterType {
     fn from(hostnames: Vec<String>) -> Self {
-        if hostnames.len() == 1 && hostnames[0].contains(&CAPELLA_SRV_SUFFIX.to_string()) {
+        if hostnames.len() == 1 && (hostnames[0].contains(&CAPELLA_SRV_SUFFIX.to_string())) {
```
chvck, line 18: I'm assuming this code only gets run if a cluster_type isn't specified in the config?

chvck (approved):

src/remote_cluster.rs
Westwooo: It always gets run but the result is ignored if the cluster_type has been specified

PR #356, opened 2024-07-24
Fix flaky import_sample_already_loaded test

Westwooo:


tests/buckets_sample.rs
```diff
@@ -14,7 +15,7 @@ fn import_sample() {
         assert_eq!(r#""success""#, json["status"].to_string());
 
         // Wait for the bucket to finish being created
-        thread::sleep(time::Duration::from_millis(5000));
+        thread::sleep(time::Duration::from_secs(10));
```
chvck, line 18: We should really poll here rather than using a sleep which a) is guaranteed to slow down the tests by at least the duration specified, and b) is probably still going to be flakey in some environments.

chvck (changes requested):

chvck (approved):

PR #376, opened 2024-08-05
Allow ip command

Westwooo:


src/cli/allow_ip.rs
```diff
@@ -0,0 +1,147 @@
...
+            .category(Category::Custom("couchbase".to_string()))
+            .optional(
+                "address",
+                SyntaxShape::String,
+                "ip address to allow access to the cluster",
+            )
+    }
+
+    fn usage(&self) -> &str {
+        "Whitelists an IP address on a Capella cluster"
```
chvck, line 39: "Adds an IP address to the allowlist on a Capella cluster" maybe? I think we've moved away from using whitelist?

src/cli/allow_ip.rs
```diff
@@ -0,0 +1,147 @@
...
+            deadline,
+            ctrl_c,
+        )
+        .map_err(|e| client_error_to_shell_error(e, span))?;
+
+    Ok(PipelineData::empty())
+}
+
+fn format_ip_address(ip_address: String) -> String {
+    if !ip_address.ends_with("/32") {
```
chvck, line 142: ip address can't be any other format can it? Just checking that we aren't blocking some usage.

Westwooo: The addresses required a subnet mask, and if not given the UI defaults to try and use '/32', so I've changed this to behave the same.

src/cli/allow_ip.rs
Westwooo: Done!

ldoguin:
Fails in loop if the current project is not set.

Westwooo:
> Fails in loop if the current project is not set.

That's expected, unless passing in a projects with the --project flag

Matt-Woz (approved):
LGTM 👍

PR #377, opened 2024-08-05
Register cluster to env on creation

Westwooo:
When we craete a cluster with the shell we should automatically register it to cb-env for future use

### DIFF
--- a/src/cli/clusters_create.rs
+++ b/src/cli/clusters_create.rs
@@ -6,8 +6,14 @@ use std::ops::Add;
 use std::sync::{Arc, Mutex};
 use tokio::time::Instant;
 
+use crate::cli::created_cluster_not_registered;
 use crate::cli::error::{client_error_to_shell_error, serialize_error};
 use crate::cli::util::{find_org_id, find_project_id};
+use crate::client::RustTlsConfig;
+use crate::config::DEFAULT_KV_BATCH_SIZE;
+use crate::remote_cluster::{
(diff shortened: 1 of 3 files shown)


src/cli/error.rs
```diff
@@ -259,6 +262,9 @@ impl From<CBShellError> for ShellError {
             CBShellError::LLMApiKeyMissing {provider} => {
                 spanned_shell_error(format!("api_key required to use {} models", provider), Some("Define an api_key in the config/credentials file".to_string()), None)
             }
+            CBShellError::CreatedClusterNotRegistered {inner_error} => {
+                spanned_shell_error(format!("Cluster created but not registered: {}", inner_error), None, None)
```
chvck, line 266: I think this should be updated to add a bit more information about what "not registered" means.

src/cli/clusters_create.rs
```diff
@@ -152,9 +161,40 @@ fn clusters_create(
...
+        .get_cluster(definition.name(), org_id, project_id, deadline, ctrl_c)
+        .map_err(|e| created_cluster_not_registered(e.message()))?;
+
+    let tls_config =
+        RustTlsConfig::new(true, None).map_err(|e| created_cluster_not_registered(e.message()))?;
+
+    let cluster = RemoteCluster::new(
+        RemoteClusterResources {
+            hostnames: vec![cluster_json.connection_string()],
+            username: "".to_string(),
```
chvck, line 183: Are we not able to get all this info from the definition or something? I'm not sure how much value there is registering the cluster if it's still not actually usable.

PR #382, opened 2024-08-06
Improve credentials create

Westwooo:
This does a few improvements to credentials create:

1) Allows users to specify the privileges of the creds using the --read and --write flags
2) Allows control over whether or not the newly created creds are set to be used by the active cluster
3) If the creds are set for use by the active cluster then we re-build the cluster client. This is needed otherwise when executing data ops against the cluster the new creds would not be used.

### DIFF
--- a/src/cli/credentials_create.rs
+++ b/src/cli/credentials_create.rs
@@ -3,6 +3,7 @@ use crate::cli::{client_error_to_shell_error, no_active_cluster_error};
 use crate::client::cloud_json::CredentialsCreateRequest;
 use crate::read_input;
 use crate::state::State;
+use nu_engine::CallExt;
 use nu_protocol::ast::Call;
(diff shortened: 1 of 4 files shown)


src/cli/credentials_create.rs
```diff
@@ -27,7 +28,15 @@ impl Command for crate::cli::CredentialsCreate {
...
 
     fn signature(&self) -> Signature {
-        Signature::build("credentials create").category(Category::Custom("couchbase".to_string()))
+        Signature::build("credentials create")
+            .category(Category::Custom("couchbase".to_string()))
+            .switch("read", "enable read access", None)
+            .switch("write", "enable write access", None)
+            .switch(
+                "registered",
+                "create credentials with the username/password the cluster was registered with",
```
chvck, line 37: "the active cluster" might be a bit clearer here?

chvck (approved):

PR #386, opened 2024-08-09
Remove vector search unwraps

Westwooo:


src/cli/vector_search.rs
```diff
@@ -134,41 +134,34 @@ fn run(
                 let id = rec.get("id").unwrap().as_str().unwrap();
```
chvck, line 134: Won't this line and the line about content both panic on unexpected types?

Westwooo: This is inside an if: 
```
if rec.contains("id") && rec.contains("content") {
```

chvck (approved):

PR #400, opened 2024-08-21
Add flag to pass in project to run clusters commands against

Westwooo:
At the moment all the clusters commands run against the active project. I'd be nice to have a flag to override this behaviour for easier management of clusters across various projects.

### DIFF
--- a/src/cli/clusters_get.rs
+++ b/src/cli/clusters_get.rs
@@ -36,6 +36,12 @@ impl Command for ClustersGet {
                 "the Capella organization to use",
                 None,
             )
+            .named(
+                "project",
+                SyntaxShape::String,
+                "the Capella project to use",
+                None,
+            )
             .category(Category::Custom("couchbase".to_string()))
     }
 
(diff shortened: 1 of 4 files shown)


src/cli/clusters_get.rs
```diff
@@ -75,13 +81,20 @@ fn clusters_get(
     } else {
         guard.active_capella_org()
     }?;
+
+    let project = if let Some(proj) = call.get_flag(engine_state, stack, "project")? {
+        proj
+    } else {
+        guard.active_project()?
+    };
```
mattyhall, line 89: Matter of taste but I think this could also be written

```suggestion
    call.get_flag(engine_state, stack, "project")?.map_or_else(|| guard.active_project()?, Ok)?; 
```

I think I prefer yours :P, but it is [in the docs as an example](https://doc.rust-lang.org/stable/std/option/enum.Option.html#method.map_or_else)

Westwooo: I can't argue with the docs, so have done

Peter-Searby (approved):

src/cli/clusters_get.rs
ldoguin: I asked my favorite rustacean and he said he prefers map_or_else when returning only one thing. And prefer the if let when there are possible side effects.

Westwooo: Sounds like a good rule of thumb

PR #401, opened 2024-08-22
Allow username and password to be passed into register with flags

Westwooo:
By default the `cb-env register` command prompts for the user name and password to be entered after running the command, allowing the password to be hidden. We should support passing the username and password in using flags to allow registering multiple clusters without having to repeatedly enter the same username and password.


src/cli/util.rs
```diff
@@ -444,21 +444,32 @@ pub fn get_active_cluster<'a>(
...
+    password_flag: Option<String>,
+) -> Result<(String, String), ShellError> {
+    let username = user_flag.map_or_else(
+        || {
+            println!("Please enter username:");
+            match read_input() {
+                Some(user) => Ok(user),
+                None => {
+                    return Err(generic_error("Username required", None, None));
+                }
```
mattyhall, line 458: Could use [`Option::ok_or_else`](https://doc.rust-lang.org/std/option/enum.Option.html#method.ok_or_else) here

src/cli/util.rs
```diff
@@ -444,21 +444,32 @@ pub fn get_active_cluster<'a>(
...
+            }
+        },
+        |u| Ok(u),
```
mattyhall, line 461: Just `Ok` (without the closure) should suffice I think

mattyhall (commented):
LGTM

src/cli/util.rs
Westwooo: Done!

src/cli/cbenv_register.rs
```diff
@@ -42,6 +42,18 @@ impl Command for CbEnvRegister {
                 SyntaxShape::String,
                 "the connection string to use for this cluster",
             )
+            .named(
+                "username",
+                SyntaxShape::String,
+                "the username to use for the registered cluster",
+                None,
```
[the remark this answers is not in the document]
Westwooo: We haven't really in the past, but there is no reason I'm aware of. I'll add some short options for password and username since shortening them is pretty standard.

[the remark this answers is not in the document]
Peter-Searby: Did you mean to remove the shortened fields again?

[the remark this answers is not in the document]
Westwooo: Yeah I made them `-u` and `-p` nut then thought that we'd want to reserve p for the shortening of the project flag since this will be more common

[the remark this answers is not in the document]
Peter-Searby: Makes sense

src/cli/util.rs
```diff
@@ -444,21 +444,35 @@ pub fn get_active_cluster<'a>(
...
+            println!("Please enter username:");
+            read_input().ok_or_else(|| generic_error("Username required", None, None))
+        },
+        Ok,
+    )?;
+
+    let password = password_flag.map_or_else(
+        || match rpassword::prompt_password("Password: ") {
+            Ok(p) => {
+                if p.is_empty() {
```
Peter-Searby, line 462: Doesn't look like the original code had this check, is it needed?
I assume there could be cases where we don't want to include a password at all (e.g. when using certs). Does this check get avoided for those cases?

Westwooo: The original code didn't have this check as I thought that `prompt_password` returned an error on am empty password, but it does not. 

Even when using certs cbshell needs the username and password to perform any doc ops. Also when we register a cluster through the config file we return an error if there is no username and password, so I think it should be the same when registering a cluster through this command. 

The other place this is used is when creating credentials, where we obviously need a username and password. I think if there was something that used wither certs or username/password it'd be the responsibility of the caller to call this func if they require a username and password, and don't if a cert has been supplied.

Peter-Searby: Thanks for the context, makes sense.

Peter-Searby (approved):

Issue #409, opened 2024-08-23
Add username and password flags to credentials create


Westwooo:
We should add these flags to allow creating credentials against multiple clusters in a loop. Otherwise the user would have to keep entering the username and password manually.

PR #410, opened 2024-08-23
Add username and password flags to credentials create

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/409

### DIFF
--- a/src/cli/credentials_create.rs
+++ b/src/cli/credentials_create.rs
@@ -7,6 +7,7 @@ use crate::state::State;
 use nu_engine::CallExt;
 use nu_protocol::ast::Call;
 use nu_protocol::engine::{Command, EngineState, Stack};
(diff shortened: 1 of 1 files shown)


src/cli/credentials_create.rs
```diff
@@ -33,6 +34,18 @@ impl Command for crate::cli::CredentialsCreate {
...
+            .named(
+                "username",
+                SyntaxShape::String,
+                "the username to use for the registered cluster",
+                None,
+            )
+            .named(
+                "password",
+                SyntaxShape::String,
+                "the password to use with the registered cluster",
```
hummusonrails, line 46: I would maintain parallelism between the two instructions so using **for** in both instances

```suggestion
                "the password to use for the registered cluster",
```

Matt-Woz (approved):

Issue #404, opened 2024-08-22
Support default projects in config file


Westwooo:
When registered to cbshell a cluster can have a default org. This is the capella-organization that will be used when the cluster is active, the org-if is required for any v4 API calls. We should support defining a default project in a similar way, so that if a used has many clusters in different projects they can register them with default projects and won't have to keep running 'cb-env project' after 'cb-env cluster'.

We should also add a flag for `default-project` when registering a cluster with `cb-env register`

PR #414, opened 2024-08-27
Add support for default project for registered clusters

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/404

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -385,6 +385,7 @@ impl ClusterConfigBuilder {
             tls: self.tls.unwrap_or_default(),
             kv_batch_size: None,
             capella_org: None,
+            project: None,
             cluster_type: None,
         }
     }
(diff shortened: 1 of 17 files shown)


src/config.rs
```diff
@@ -421,6 +422,9 @@ pub struct ClusterConfig {
     ))]
     capella_org: Option<String>,
 
+    #[serde(default)]
```
Peter-Searby, line 425: I'm assuming this just means that when deserialising it will default to specifying it as an empty string?

Matt-Woz: If unspecified, serde will deserialise it into a `None` (And `None` is serialised into a JSON `null`)

Peter-Searby: Thanks @Matt-Woz 
@Westwooo I don't really get why we want this to be null when deserialised, but others, such as cluster_type would instead presumably just not have a value?

Westwooo: Good point. cluster_type and project would both deserialise into None when missing since they are options, so have removed the unnecessary `#[serde(default)]` attribute, also have removed it from capella_organization.

Peter-Searby (approved):

PR #428, opened 2024-09-02
Similarity search recipe

Westwooo:


docs/recipes/similarity_search.adoc
```diff
@@ -0,0 +1,125 @@
...
+│ 4 │ Sydney G. Walton Square   │                                       │ Small (one square block), well maintained park/square in the heart of the city, located right beside the Financial District.  Tai Chi practitioners exercise here in    │
+│   │                           │                                       │ the early morning hours.                                                                                                                                                │
+╰───┴───────────────────────────┴───────────────────────────────────────┴─────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
+```
+
+Here we have done another similarity search using the same index, but our source vector is the result of embedding the phrase "physical exercise".
+One important detail to remeber is that the embedding generated from `vector enrich-text` must have the same dimension as those over which the index was created, otherwise `vector search` will return no results.
```
ingenthr, line 124: "dimensions" instead of "dimension"?

docs/recipes/similarity_search.adoc
```diff
@@ -0,0 +1,125 @@
...
+See <<_vector_enrich-text, vector enrich-text>> for how to specify the dimension of the generated embeddings.
```
ingenthr, line 125: Hm, maybe it was intentional.

ingenthr (approved):
To me, it should read as 'dimenions', since it's always more than one, but I'll defer to your expertise in both what's canonical for this domain and what's correct English.

Westwooo:
> defer

I agree that dimensions makes more sense in the english sense, but I always see dimension used. I understand it as it's because you're taking about the dimension of the vector space to which the vector belongs.

PR #429, opened 2024-09-02
Managing multiple clusters recipe

Westwooo:


docs/recipes/managing_multiple_clusters.adoc
```diff
@@ -0,0 +1,164 @@
...
+👤 Charlie 🏠 obligingfaronmoller in ☁️ default._default._default
+> cb-env managed
+╭───┬────────┬───────┬────────────────────────┬───────────────┬──────────────────────┬─────────────────╮
+│ # │ active │  tls  │       identifier       │   username    │ capella_organization │     project     │
+├───┼────────┼───────┼────────────────────────┼───────────────┼──────────────────────┼─────────────────┤
+│ 0 │ false  │ true  │ meekjoycekreynolds     │ Administrator │ my-org               │ CBShell Testing │
+│ 1 │ false  │ false │ local                  │ Administrator │                      │                 │
+│ 2 │ false  │ true  │ bronzegeraldjaysussman │ Administrator │ my-org               │ CBShell Testing │
+│ 3 │ true   │ true  │ obligingfaronmoller    │ Administrator │ my-org               │ CBShell Testing │
+╰───┴────────┴───────┴────────────────────────┴───────────────┴──────────────────────┴─────────────────╯
```
ingenthr, line 16: Minor improvement if you want, give them identifiers that match a narrative.  like "localdev" and "systemtest" and "prod" and "ci".

docs/recipes/managing_multiple_clusters.adoc
```diff
@@ -0,0 +1,164 @@
...
+│ 7 │ obligingfaronmoller    │    13.6 GiB │
+│ 8 │ obligingfaronmoller    │    13.6 GiB │
+│ 9 │ meekjoycekreynolds     │    14.4 GiB │
+╰───┴────────────────────────┴─────────────╯
+```
+
+We do this by iterating over each node and https://www.nushell.sh/commands/docs/update.html[updating] the value in the `memory_free` column by multiplying the current value by nushell's inbuilt https://www.nushell.sh/book/types_of_data.html#file-sizes[File Size] datatype.
+
+While it is somewhat useful to know the free memory that each cluster has, it'd be more useful for our healthcheck to know the memory utilization for each cluster.
+Unfortunately the info returned by `nodes` does not include the memory utilization, however there are two columns that can be used to calculate this: `memory_free` and `memory_total`.
```
ingenthr, line 90: Again, quite cool, but it might be neat to fit it into a narrative.  LIke it's useful to find a test cluster with space to create a bucket, yadda, yadda.

ingenthr (approved):

docs/recipes/managing_multiple_clusters.adoc
Westwooo: The narrative started off as performing health checks on the cluster, then once the health is understood we can use the resource info to choose where to allocate a bucket. I've changed this to link it more heavily to the health check then I have the bit about the bucket later on: 

```
Now that we understand the resources being used by our cluster we can use this information to aide us when deploying a new bucket.
Imagine that we want to create a 1GB bucket on any one of our clusters.
```

docs/recipes/managing_multiple_clusters.adoc
Westwooo: Done

PR #430, opened 2024-09-02
Register healthy Capella clusters recipe

Westwooo:


docs/recipes.adoc
```diff
@@ -5,6 +5,8 @@
 Welcome to the recipes section of the Couchbase Shell `cbsh` documentation.
 Here you can find some examples of the more complicated tasks that can be performed using `cbsh`.
```
ingenthr, line 6: I know this wasn't changed here, but maybe…

"Here you can find some examples of how powerful tasks can be performed in a set of pipelined statements using `cbsh`." or something like that?

docs/recipes/register_cluster.adoc
```diff
@@ -0,0 +1,81 @@
+== Register clusters
+
+Before CBShell can be used to perform data operations on a cluster, the cluster must be registered with the shell.
+Cluster registration can be done through the https://couchbase.sh/docs/#_the_config_dotfiles[config file], or using https://couchbase.sh/docs/#_cb_env_register[cb-env register].
+This recipe shows how to register all the healthy clusters in the active https://couchbase.sh/docs/#_the_config_dotfiles[registered organization].
+
+Firstly we need to get a list of all the healthy clusters, along with the projects to which they belong:
```
ingenthr, line 7: en_US would be "First, " rather than "Firstly.

docs/recipes/register_cluster.adoc
```diff
@@ -0,0 +1,81 @@
...
+```
+
+This creates read/write credentials on all of the registered clusters that match the https://couchbase.sh/docs/#_the_clusters_flag[--clusters] regex using the username and password with which the clusters were registered to the shell.
+Finally we need the cluster to allow requests from our IP address, this can be done as follows:
+
+```
+> curl ipconfig.me | allow ip --clusters .*
+```
+
+The curl command returns out IP address, which we then pipe into allow ip` and this is run against all of the registered clusters again.
```
ingenthr, line 77: "our IP address"

ingenthr (changes requested):
typo!

docs/recipes/register_cluster.adoc
Westwooo: Chagned

docs/recipes/register_cluster.adoc
Westwooo: Done

PR #436, opened 2024-09-04
Add recipe for moving buckets/scopes/collections and data

Westwooo:


docs/recipes/moving_data.adoc
```diff
@@ -0,0 +1,213 @@
...
+```
+
+The first thing to do is to recreate all of the buckets that we have on the `local` cluster on the `remote` cluster:
+
+```
+> buckets | each {|in| buckets create $in.name ($in.ram_quota / 1MB | into int)  --clusters remote --replicas 1}
+```
+
+Here we simply get all of the buckets, then iterate over the list with https://www.nushell.sh/commands/docs/each.html[each] and create buckets with the same name and ram quota, specifying the `remote` cluster with the https://couchbase.sh/docs/#_the_clusters_flag[--clusters] flag.
+Since the value for the ram quote is returned in bytes from `buckets` we convert it to MiB bby dividing by nushells 1MB https://www.nushell.sh/book/types_of_data.html#file-sizes[FileSize] datatype.
```
ingenthr, line 38: bby?

docs/recipes/moving_data.adoc
```diff
@@ -0,0 +1,213 @@
...
+╭───┬─────────┬───────────────┬───────────┬──────────┬──────────────────────┬───────────┬───────────────┬───────┬────────────╮
+│ # │ cluster │     name      │   type    │ replicas │ min_durability_level │ ram_quota │ flush_enabled │ cloud │ max_expiry │
+├───┼─────────┼───────────────┼───────────┼──────────┼──────────────────────┼───────────┼───────────────┼───────┼────────────┤
+│ 0 │ remote  │ beer-sample   │ couchbase │        1 │ none                 │ 209.0 MiB │ false         │ true  │          0 │
+│ 1 │ remote  │ travel-sample │ couchbase │        1 │ none                 │ 209.0 MiB │ false         │ true  │          0 │
+╰───┴─────────┴───────────────┴───────────┴──────────┴──────────────────────┴───────────┴───────────────┴───────┴────────────╯
+```
+
+Next we need to create all of the scopes within these buckets.
+Firstly we get all the buckets again on the `local` cluster, then for each of the buckets we get the scopes:
```
ingenthr, line 54: en_US

ingenthr (changes requested):

docs/recipes/moving_data.adoc
Westwooo: Corrected

docs/recipes/moving_data.adoc
Westwooo: Uncorrected

ingenthr (approved):

Issue #440, opened 2024-09-05
Change default replica number based on number of nodes


Westwooo:
Currently when a user creates a bucket the `num_replicas` defaults to 0. This returns an error when run against multi-node Capella clusters which support a minimum `num_replicas` of 1. If num_replicas has not been explicitly set by the user we need to check the number of nodes on the cluster and set `num_replicas` as appropriate.

PR #441, opened 2024-09-05
Set num_replicas to 0 for buckets created on single node clusters

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/440

The `unwrap()`s in:

- buckets_get is fine, since num_replicas will always be set when fetching the settings, it's only going to be None when creating a bucket
- cloud_json.rs/total_nodes is fine since we only call it on fetched json clusters, which will always have this set

### DIFF
--- a/src/cli/buckets_create.rs
+++ b/src/cli/buckets_create.rs
@@ -1,4 +1,6 @@
-use crate::cli::buckets_builder::{BucketSettingsBuilder, BucketType, DurabilityLevel};
+use crate::cli::buckets_builder::{
+    BucketSettings, BucketSettingsBuilder, BucketType, DurabilityLevel,
+};
 use crate::cli::error::{
     client_error_to_shell_error, generic_error, serialize_error, unexpected_status_code_error,
 };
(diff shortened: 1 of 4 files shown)


src/cli/buckets_create.rs
```diff
@@ -247,12 +248,25 @@ pub fn create_capella_bucket(
         project_id.clone(),
     )?;
 
+    // If we leave num_replicas as empty the API always sets the default to 1 which will fail for
```
Matt-Woz, line 251: What is meant by `the API always sets the default to 1`? I ask because in SDK-land we try to keep the SDK unopinionated & leave unspecified fields unset in the request for the server to decide on sensible defaults, which is fine in the single-node bucket_create case, so I'm curious why cbsh is different here

Westwooo: I was trying to do the same, and just leave it blank to let the V4 API supply a sensible default. However the default is sensible only for multi-node clusters since it is 1 which will return an error for single node clusters.

Matt-Woz: Ah I see, that's an odd choice for the default. I wonder if in that case we should only override num_replicas `if total_nodes() == 1`, else keep it unset?

Westwooo: Yes we should now you mention it, have made that change

Issue #487, opened 2024-09-19
Support configuration of Capella API endpoint


Westwooo:
Currently we check the suffix of the cluster endpoint to determine if the cluster is running in the cloud. This works fine for clusters running in prod, but the dev environments have different endings to the connstr. This issue was addresses by the ability to manually specify the cluster type in the config file. 

However the dev environments also have a different endpoint for the Capella API, so we need to be able to configure this through the config file also.

PR #498, opened 2024-09-24
Support configuration of api endpoint for Capella organization

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/487

### DIFF
--- a/src/cli/cbenv_register.rs
+++ b/src/cli/cbenv_register.rs
@@ -230,6 +230,7 @@ pub fn update_config_file(guard: &mut MutexGuard<State>, span: Span) -> Result<(
             c.access_key(),
             Some(c.timeout()),
             c.default_project(),
+            None,
         ))
     }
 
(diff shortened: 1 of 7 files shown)


src/cli/cbenv_register.rs
```diff
@@ -230,6 +230,7 @@ pub fn update_config_file(guard: &mut MutexGuard<State>, span: Span) -> Result<(
             c.access_key(),
             Some(c.timeout()),
             c.default_project(),
+            None,
```
Peter-Searby, line 233: Why is this None?
Presumably we would want to save the api_endpoint to the config file?

Westwooo: I don't think that we do. Since either the endpoint will be the default, so adding the default endpoint to the file will just cause confusion. Or it will not be the default, in which case it will already be in the config file.

Peter-Searby: Ah, I see. Since it's only configurable through the config file, it doesn't make sense to update the config file.

Peter-Searby (approved):

Issue #501, opened 2024-09-26
Links between docs and recipes not working


Westwooo:
There are links between the docs and recipes pages that should lead to a specific anchor on the external page, but they just go to the top of the page. This is because of the was we are embedding html files, so we need to manually check for an anchor in the url and scroll there if it present.

PR #502, opened 2024-09-26
Add auto scrolling js

Westwooo:
Part of the solution to: https://github.com/couchbaselabs/couchbase-shell/issues/501


Peter-Searby (approved):
It would be nice to not have such a hacky solution, and it would be nice not to have the issue where scrolling doesn't always go to the exactly correct position.
However, since it is better to get near the anchor point than to just end up at the top of the page entirely, I'm approving this change.
However, please file a tracking issue for the slightly-off scrolling

PR #525, opened 2024-10-08
Add support for listing columnar databases

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/522


src/cli/columnar_databases.rs
```diff
@@ -0,0 +1,114 @@
...
+}
+
+impl Command for ColumnarDatabases {
+    fn name(&self) -> &str {
+        "columnar databases"
+    }
+
+    fn signature(&self) -> Signature {
+        Signature::build("columnar databases")
+            .switch("with-meta", "Includes related metadata in the result", None)
```
chvck, line 31: Is this being used?

Westwooo: It was not, good spot. Have removed now.

chvck (approved):

Issue #529, opened 2024-10-09
Redact secrets in debug logs


Westwooo:
When the configs are printed when the log level is debug, we print passwords and API keys in plain text. These sensitive fields should be redacted so that users can share their logs without worrying.

PR #543, opened 2024-10-17
Redact secrets from debug logs

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/529

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -3,9 +3,10 @@ use crate::state::Provider;
 use log::debug;
 use log::error;
 use serde::{Deserialize, Serialize};
-use std::fs;
+use std::fmt::Debug;
 use std::path::PathBuf;
 use std::time::Duration;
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -324,6 +325,18 @@ impl LLMConfig {
     }
 }
 
+impl Debug for LLMConfig {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("OrganizationCredentials")
+            .field("identifier", &self.identifier)
+            .field("api_key", &"****")
```
Matt-Woz, line 332: Should we use `.map(|_| "****".to_string()` on the`api_key`, given it's an `Option`?

Westwooo: Yes we should, I forgot it was optional

Matt-Woz (approved):

Issue #535, opened 2024-10-11
Analytics query streaming


petrkozorezov:
I have an analytics dataset with >100m rows and want to make a query over it and dump to parquet or csv for further analysis. Is it possible to do without intermediate storing in memory?

Westwooo:
Hello @petrkozorezov, the easiest way to do this would be: 

```
analytics "your analytics query here" | save results.csv
```

I believe that save will stream the data without collection.

petrkozorezov:
It doesn't look like a streaming. It download first, then eats a lot of memory, and then writes everything to disk.
In my case it works with 5 columns and 60m rows (using around 60Gb ram), but doesn't with 80m (because of OOM killer) (the same query but with different 'limit').

Westwooo:
You're correct, I think it could be possible to add support for streaming to the analytics command, I'll leave this issue up as a request for that feature.

petrkozorezov:
related #162

Westwooo:
This will be in the next release, please let us know if you run into any issues with is @petrkozorezov

petrkozorezov:
> This will be in the next release, please let us know if you run into any issues with is @petrkozorezov

I've built the latest version from master.
Type is stream now, but behavior hasn't changed.

was:
```
> analytics "SELECT META().id FROM my_table LIMIT 10" | describe
table<id: string, cluster: string>
```

now:
```
> analytics "SELECT META().id FROM my_table LIMIT 10" | describe
table<id: string, cluster: string> (stream)
```

I tried a query, but it ate all memory (>120Gb) and then stuck.
```
> analytics "SELECT my_fields FROM my_table" | to csv | save --raw -f test.csv
```

Westwooo:
> > This will be in the next release, please let us know if you run into any issues with is @petrkozorezov
> 
> I've built the latest version from master. Type is stream now, but behavior hasn't changed.
> 
> was:
> 
> ```
> > analytics "SELECT META().id FROM my_table LIMIT 10" | describe
> table<id: string, cluster: string>
> ```
> 
> now:
> 
> ```
> > analytics "SELECT META().id FROM my_table LIMIT 10" | describe
> table<id: string, cluster: string> (stream)
> ```
> 
> I tried a query, but it ate all memory (>120Gb) and then stuck.
> 
> ```
> > analytics "SELECT my_fields FROM my_table" | to csv | save --raw -f test.csv
> ```

Can you try without the `to csv` in the middle?

[...]

PR #551, opened 2024-10-29
Analytics streaming

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/535


src/cli/analytics.rs
```diff
@@ -77,51 +86,148 @@ impl Command for Analytics {
     }
 }
 
+pub struct AnalyticsStream {
+    span: Span,
+    streams: StreamMap<String, RawJsonRowStreamer>,
+    rt: Arc<Runtime>,
```
Westwooo, line 92: This field never gets read but is required as otherwise the runtime is closed at the end of the function before we can read from the streams that were created in it. This was the runtime's lifetime is the same as the streams

src/cli/analytics.rs
```diff
@@ -77,51 +84,148 @@ impl Command for Analytics {
...
     let cluster_identifiers = cluster_identifiers_from(engine_state, stack, &state, call, true)?;
 
     let guard = state.lock().unwrap();
 
+    let ctrl_c = engine_state.ctrlc.as_ref().unwrap().clone();
+    let statement: String = call.req(engine_state, stack, 0)?;
+
     let scope: Option<String> = call.get_flag(engine_state, stack, "scope")?;
-    let with_meta = call.has_flag(engine_state, stack, "with-meta")?;
+    let _with_meta = call.has_flag(engine_state, stack, "with-meta")?;
```
chvck, line 141: Why is with_meta no longer used? If we don't want it to be used then should we remove it from the command?

src/cli/analytics.rs
```diff
@@ -77,51 +84,148 @@ impl Command for Analytics {
...
+    fn next(&mut self) -> Option<Self::Item> {
+        if let Some((cluster, result)) = self
+            .rt
+            .clone()
+            .block_on(async { self.streams.next().await })
+        {
+            let bytes = match result {
+                Ok(r) => r,
+                Err(e) => {
+                    return Some(Value::Error {
```
chvck, line 107: Does this mean that the iterator is going to try to continue even after an error?

src/cli/analytics.rs
```diff
@@ -77,51 +84,148 @@ impl Command for Analytics {
...
+
+    Ok(PipelineData::from(ListStream::new(
+        result_stream,
+        span,
+        Some(ctrl_c.clone()),
+    )))
+}
+
+pub async fn send_analytics_query_stream(
+    client: HTTPClient,
```
chvck, line 188: Why are we bypassing the standard interface we use for sending requests?

src/client/http_client.rs
```diff
@@ -336,6 +338,61 @@ impl HTTPClient {
         })
     }
 
+    pub async fn analytics_query_stream_request(
```
chvck, line 341: Can we just use analytics_query_request and always return a streaming response? It should be fairly to add a helper to consume the stream at a higher level for things like index management.

src/client/http_handler.rs
```diff
@@ -206,4 +208,102 @@ impl HTTPHandler {
...
+            impl Stream<Item = Result<Bytes, reqwest::Error>> + Sized,
+            u16,
+        ),
+        ClientError,
+    > {
+        self.http_do_stream(uri, HttpVerb::Post, payload, headers, deadline, ctrl_c)
+            .await
+    }
+
+    pub async fn http_do_stream(
```
chvck, line 231: This is another case where it seems like we don't need both http_do and http_do_stream and http_do just streams anyway. Either post, put, get etc... should also return a stream or we leave post_stream in and the others just consume the stream before returning.

src/cli/analytics.rs
Westwooo: No when an error is returned the iterator stops being read, I altered the code to always return an error then ran: 

```
analytics "FROM range(0,1000000) AS r SELECT repeat("testing", 10) AS x1, repeat("testing", 10) AS x2"
```

And only one error was returned, then the stream stopped, instead of seeing 1000000 errors.

src/cli/analytics.rs
```diff
@@ -77,51 +85,103 @@ impl Command for Analytics {
...
             span,
-            with_meta,
-            true,
-        )?);
-    }
+            rt.clone(),
+        )?;
+
+        let json_stream = JsonRowStream::new(stream);
+        let mut json_streamer = RawJsonRowStreamer::new(json_stream, "".to_string());
```
chvck, line 164: This is a bit suspicious, why is there no rows attrib?

src/cli/analytics.rs
```diff
@@ -77,51 +85,103 @@ impl Command for Analytics {
...
+
+        rt.block_on(async {
+            // Read prelude and signature
+            json_streamer.read_prelude().await.unwrap();
+            json_streamer.read_row().await.unwrap();
```
chvck, line 169: i suspect that these read_row calls are needed because of no specifying a rows attrib, I don't think they should be needed.

src/cli/analytics.rs
```diff
@@ -77,51 +85,103 @@ impl Command for Analytics {
...
+    let result_stream = AnalyticsStream { streams, span, rt };
+
+    Ok(PipelineData::from(ListStream::new(
+        result_stream,
+        span,
+        Some(ctrl_c.clone()),
+    )))
 }
 
 pub fn send_analytics_query(
```
chvck, line 187: Having send and do is really confusing, should we rename these to make intent clear?

src/client/http_client.rs
```diff
@@ -332,17 +335,17 @@ impl HTTPClient {
...
-    pub fn analytics_query_request(
+    pub fn analytics_query_stream_request(
         &self,
         request: AnalyticsQueryRequest,
         deadline: Instant,
         ctrl_c: Arc<AtomicBool>,
-    ) -> Result<HttpResponse, ClientError> {
-        let rt = Runtime::new().unwrap();
+        rt: Arc<Runtime>,
+    ) -> Result<(Pin<Box<ResultsStream>>, u16), ClientError> {
```
chvck, line 348: This is a bit confusing to me, why doesn't this return HttpStreamResponse?

src/client/http_handler.rs
```diff
@@ -32,20 +41,53 @@ impl HttpVerb {
...
+
+    pub fn content(&self) -> &str {
+        &self.content
+    }
+
+    pub fn status(&self) -> u16 {
+        self.status
+    }
+
+    // pub fn endpoint(&self) -> Endpoint {
```
chvck, line 59: What was this used by that doesn't need it anymore?

src/client/http_handler.rs
```diff
@@ -183,27 +225,47 @@ impl HTTPHandler {
...
+    pub(crate) async fn http_post(
         &self,
         uri: &str,
         payload: Option<Vec<u8>>,
         headers: HashMap<&str, &str>,
         deadline: Instant,
         ctrl_c: Arc<AtomicBool>,
-    ) -> Result<(String, u16), ClientError> {
-        self.http_do(uri, HttpVerb::Put, payload, headers, deadline, ctrl_c)
+    ) -> Result<(Pin<Box<ResultsStream>>, u16), ClientError> {
```
chvck, line 247: It might be nice to type alias this Pin<Box<ResultStream>> type

src/cli/analytics.rs
Westwooo: I wasn't sure what the row_attrib was/should be and tried a few things but couldn't get it to work

src/client/http_handler.rs
Westwooo: The methods that now return a HttpStreamingResponse used to use this. Now HttpResponse is only used by the Capella methods which never populated/read this field.

src/client/http_client.rs
Westwooo: That was my mistake, corrected now

src/cli/analytics.rs
Westwooo: Good point, I've taken send out of do, and made do `read_analytics_response`

src/client/http_handler.rs
Westwooo: Done

src/cli/analytics.rs
Westwooo: I wasn't sure what is should be, have checked the usage in couchbase-rs and is now working without the hacky reading of the first couple of rows.

src/cli/analytics.rs
```diff
@@ -77,51 +84,99 @@ impl Command for Analytics {
...
 
-    Ok(Value::List {
-        vals: results,
-        internal_span: span,
+        let json_stream = JsonRowStream::new(resp.stream());
+        let mut json_streamer = RawJsonRowStreamer::new(json_stream, "results");
+
+        rt.block_on(async {
+            // Read prelude so rows are ready for reading
+            json_streamer.read_prelude().await.unwrap();
```
chvck, line 167: Is this going to panic if the user hits ctrl-c?

src/client/http_client.rs
```diff
@@ -238,9 +238,9 @@ impl HTTPClient {
         request: ManagementRequest,
         deadline: Instant,
         ctrl_c: Arc<AtomicBool>,
-    ) -> Result<HttpResponse, ClientError> {
-        let rt = Runtime::new().unwrap();
-        rt.block_on(async {
+    ) -> Result<HttpStreamResponse, ClientError> {
+        let rt = Arc::new(Runtime::new().unwrap());
+        rt.clone().block_on(async {
```
chvck, line 243: Why does this need to be cloned?

src/client/cloud.rs
```diff
@@ -166,11 +165,7 @@ impl CapellaClient {
             HttpVerb::Put => self.http_put(request.path().as_str(), request.payload(), ctrl_c)?,
         };
         // This endpoint is pretty undenyably a hack, but doesn't really matter for now.
```
chvck, line 167: NIT: comment no longer relevant

src/client/http_client.rs
chvck: For that matter why does it need to be Arc at all?

Westwooo: Needs to be an Arc so that we can clone it. And needs to be cloned as we return it from the function as part of the HttpStreamResponse. Needs to be part of the HttpStreamResponse as if we try and read the stream from a different runtime we panic.

src/cli/analytics.rs
Westwooo: Not sure, but have removed the unwrap and propagated the error.

chvck (approved):

Issue #556, opened 2024-11-12
Analytics `--with-meta` flags does not work


petrkozorezov:
```
> analytics --with-meta "select 1"
╭───┬────┬─────────╮
│ # │ $1 │ cluster │
├───┼────┼─────────┤
│ 0 │  1 │ prod    │
╰───┴────┴─────────╯
```

Westwooo:
Fixed, but the results stream is consumed when the with-meta flag is used, see PR comments for further explanation.

PR #561, opened 2024-12-06
Fix with meta flag for analytics queries

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/556

This change means that the results of an analytics query will not be streamed when the with-meta flag is used, this behaviour will need to be documented. It would be possible to have the meta returned and still stream the queries, but this has two issues: 

1) The output table would be mostly empty as all the rows will have the metadata columns but these would all be empty
2) The purpose of streaming results is to prevent reading them to memory by streaming to a file, but the format of the metadata means that it cannot be saved to file in a streaming manner.

### DIFF
--- a/src/cli/analytics.rs
+++ b/src/cli/analytics.rs
@@ -9,8 +9,10 @@ use nu_engine::command_prelude::Call;
 use nu_engine::CallExt;
 use nu_protocol::engine::{Command, EngineState, Stack};
 use nu_protocol::{
-    Category, ListStream, PipelineData, ShellError, Signature, Span, SyntaxShape, Value,
+    Category, IntoPipelineData, ListStream, PipelineData, ShellError, Signature, Span, SyntaxShape,
+    Value,
 };
+use nu_utils::SharedCow;
 use std::str::from_utf8;
 use std::sync::{Arc, Mutex};
 use tokio::runtime::Runtime;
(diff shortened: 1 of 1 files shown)


src/cli/analytics.rs
```diff
@@ -164,11 +167,60 @@ fn run(
...
+                .map_err(|e| ShellError::GenericError {
+                    error: format!("failed to read analytics query result: {}", e),
+                    msg: "".to_string(),
+                    span: None,
+                    help: None,
+                    inner: vec![],
+                })?
+            {
+                if !query_results.is_empty() {
+                    query_results.append(&mut vec![b','])
```
chvck, line 183: Why is this needed?

src/cli/analytics.rs
```diff
@@ -164,11 +167,60 @@ fn run(
...
+                .block_on(async { json_streamer.read_epilog().await })
+                .map_err(|e| ShellError::GenericError {
+                    error: format!("failed to read stream epilog: {}", e),
+                    msg: "".to_string(),
+                    span: None,
+                    help: None,
+                    inner: vec![],
+                })?;
+
+            let (start, _) = from_utf8(&epilog).unwrap().split_at(epilog.len() - 1);
```
chvck, line 198: Is there no way that we can create the nu value without manually creating a JSON value?

src/cli/analytics.rs
Westwooo: So that the results are formatted as a list

chvck (approved):

Issue #415, opened 2024-08-27
Check that `--save` works for cb-env register


Westwooo:
The save flag is supposed to update the config file with newly registered clusters. We should check that this is working correctly.

PR #565, opened 2024-12-10
Correct updating of config file when (un)registering clusters

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/415

### DIFF
--- a/src/cli/cbenv_register.rs
+++ b/src/cli/cbenv_register.rs
@@ -1,10 +1,7 @@
-use crate::config::{CapellaOrganizationConfig, ClusterConfig, ShellConfig, DEFAULT_KV_BATCH_SIZE};
-use crate::state::State;
-use std::fs;
-use std::sync::{Arc, Mutex, MutexGuard};
-
 use crate::cli::error::generic_error;
-use crate::cli::util::get_username_and_password;
+use crate::cli::util::{get_username_and_password, read_config_file, update_config_file};
+use crate::config::{ClusterConfig, DEFAULT_KV_BATCH_SIZE};
+use crate::state::State;
 use crate::{
     ClusterTimeouts, RemoteCluster, RemoteClusterResources, RemoteClusterType, RustTlsConfig,
 };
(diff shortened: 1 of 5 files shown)


src/cli/cbenv_register.rs
```diff
@@ -207,42 +205,28 @@ fn clusters_register(
...
-            c.default_project(),
+fn save_new_cluster_config(
+    guard: &mut MutexGuard<State>,
+    span: Span,
+    identifier: String,
+) -> Result<(), ShellError> {
+    let mut config = read_config_file(guard, span)?;
+    let clusters = config.clusters_mut();
+
+    if clusters.iter().any(|c| c.identifier() == identifier) {
```
chvck, line 216: This will prevent users from updating an existing cluster,is that definitely what we want?

src/cli/cbenv_register.rs
```diff
@@ -207,42 +205,28 @@ fn clusters_register(
...
-                span,
-            ));
-        }
-    };
-    let mut cluster_configs = Vec::new();
-    for (identifier, cluster) in guard.clusters() {
-        cluster_configs.push(ClusterConfig::from((identifier.clone(), cluster)))
-    }
-    let mut capella_configs = Vec::new();
-    for (identifier, c) in guard.capella_orgs() {
```
chvck, line 226: Why have we removed this?

Peter-Searby (approved):

Peter-Searby:
Ignore my review, I hadn't seen @chvck's comments (and I should have asked the same questions, as I was wondering the same)

src/cli/cbenv_register.rs
Westwooo: The Capella orgs get read and written back to the file, so we don't need to do this. With this approach the API keys always get written to the config, which we don't want if they were given in the credentials file.

src/cli/cbenv_register.rs
Westwooo: Yes since the register/unregister commands aren't intended to be used to update cluster information, only register new clusters. If you called register on an existing cluster it would error before reaching this point.

chvck (approved):

PR #583, opened 2025-02-05
Specify credential resources

Westwooo:
Addresses: https://github.com/couchbaselabs/couchbase-shell/issues/576


src/cli/credentials_create.rs
```diff
@@ -84,6 +94,44 @@ fn credentials_create(
     let read = call.has_flag(engine_state, stack, "read")?;
     let write = call.has_flag(engine_state, stack, "write")?;
     let use_registered = call.has_flag(engine_state, stack, "registered")?;
+    let bucket = call.get_flag(engine_state, stack, "bucket")?;
+    let scopes_flag: Option<Value> = call.get_flag(engine_state, stack, "scopes")?;
+
+    let scopes = if let Some(scopes) = scopes_flag {
```
Matt-Woz, line 100: NIT: I think `if let Some()` feels a bit redundant here given we're matching on `scopes` later anyway. Could we just `match` on `scopes_flag` instead? Maybe we could use [guards](https://doc.rust-lang.org/rust-by-example/flow_control/match/guard.html) for the bucket check. What do you think?

Westwooo: Good point, that's a lot neater. And thanks for the link about 'guards' wasn't aware of them :)

Matt-Woz: No problem, we could also flatten it by matching directly on e.g. `Some(Value::String { val, .. })`, but I think this is good too

Matt-Woz (approved):

PR #589, opened 2025-02-28
Initial implementation of query autocomplete

Westwooo:
This PR adds the nushell script that implements a query autocomplete feature. In this initial implementation the only supported n1ql keywords are: "FROM SELECT WHERE LIKE]. To use the feature the user needs to run: 

`use query_autocomplete *`

I'm going to look into a way of having this automatically done when cbshell is started. To generate a query you need to set an active bucket and scope using `cb-env bucket` and `cb-env scope`. Then you can do: 

`FROM `

But don't press enter use tab to display the completion options. Then you can build a query and run it by pressing enter. If you want to see the query that would be generated without running it use the `--print_query` flag. There are also tests in the module that test the completion function `fields` and the function responsible for building the final query: `format_query`. The tests for these can be run using: `fields_tests` and `format_query_tests` respectively.


query_autocomplete/fields.nu
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+    }
+
+    let $after_last_keyword = ($context | split row $last_keyword | last)
+
+    match $last_keyword {
+        FROM => {return [SELECT]}
+        SELECT => {
+            # TO DO - make this work with fields with spaces in
```
chvck, line 37: Is there an issue filed for this?

chvck (approved):
I do feel that if a user types SELECT and then hits tab we should probably just populate it with *. Possibly also logging something helpful if we can.

query_autocomplete/fields.nu
Westwooo: Have done: https://github.com/couchbaselabs/couchbase-shell/issues/590

Westwooo:
> I do feel that if a user types SELECT and then hits tab we should probably just populate it with *. Possibly also logging something helpful if we can.

I've raised: https://github.com/couchbaselabs/couchbase-shell/issues/591 so that I can have a proper think about how to handle this.

